PR #1, opened 2022-11-10
Add health check endpoint and basic project structure

liam-lloyd:
Notes on decisions:
- I started from `notification-service` and pulled things out aggressively to ensure we don't have dependencies we don't need lying around. Some of the things I pulled out for this PR I expect will be re-added shortly, like `joi`.
- In contrast to our existing backend services this project endeavors to organize code vertically rather than horizontally at the top level (so rather than `controllers` and `services` directories, we have a `health` directory, which contains the health controller and the health service). I think this does a better job of keeping related code close together, and should lead the projects directory boundaries to represent the natural sub-service boundaries along which `stela` could one day be split into microservices.

Some things intentionally not handled in this PR:
- Local development database: I think this will end up being a separate database named `permanent` on the same postgres container where the test database (`test_permanent`) runs, but this seems like a matter best addressed as part of https://permanent.atlassian.net/browse/PER-9037. For the moment, it is very much possible to develop against the test database, that database just drops all its data each test run.
- Production builds: I haven't decided how exactly this server will be run in production. Locally, it runs through `ts-node`, which is very convenient. We could use `ts-node` in prod too, but there is a slight performance cost. Alternatively, we could run from compiled Javascript files in prod, but then we have to find a way to ensure those files can find our `.sql` scripts, because `tsc` doesn't give us the option to copy those into the directory where it writes its output. This matter should be addressed as part of https://permanent.atlassian.net/browse/PER-9036.

Resolves PER-8983: Create the Typescript monolith repo (stela)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,7 @@
+[![lint](https://github.com/PermanentOrg/stela/actions/workflows/lint.yml/badge.svg)](https://github.com/PermanentOrg/stela/actions/workflows/lint.yml)
+[![unit tests](https://github.com/PermanentOrg/stela/actions/workflows/test.yml/badge.svg)](https://github.com/PermanentOrg/stela/actions/workflows/test.yml)
+[![codecov](https://codecov.io/gh/PermanentOrg/stela/branch/main/graph/badge.svg?token=4LYJGPGU57)](https://codecov.io/gh/PermanentOrg/stela)
+
 # Stela: A Monolithic Typescript Backend for Permanent.org
 
 ## Setup
--- a/README.md
+++ b/README.md
@@ -1,1 +1,43 @@
 # Stela: A Monolithic Typescript Backend for Permanent.org
+
+## Setup
+
+1. Create a `.env` file
+
+```bash
+cp .env.template .env
+```
+
+Update values as needed (see [Environment Variables](#environment-variables).
+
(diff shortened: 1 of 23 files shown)


liam-lloyd:
Changes from https://permanent.atlassian.net/browse/PER-9035 turned out to be quite small, so I've added them to this PR.

README.md
```diff
@@ -1 +1,47 @@
+[![lint](https://github.com/PermanentOrg/stela/actions/workflows/lint.yml/badge.svg)](https://github.com/PermanentOrg/stela/actions/workflows/lint.yml)
+[![unit tests](https://github.com/PermanentOrg/stela/actions/workflows/test.yml/badge.svg)](https://github.com/PermanentOrg/stela/actions/workflows/test.yml)
+[![codecov](https://codecov.io/gh/PermanentOrg/stela/branch/main/graph/badge.svg?token=4LYJGPGU57)](https://codecov.io/gh/PermanentOrg/stela)
```
liam-lloyd, line 3: If you view the README as it exists on this branch, this badge won't populate properly because it's pointed at `main`, which hasn't uploaded any coverage reports to codecov yet. I have tested that it works properly if I point it at this branch, so I'm confident it will work on `main` once this is merged.

cecilia-donnelly (approved):
Thank you for making these decisions and building our new repo!

PR #3, opened 2022-11-30
Add Dockerfile so stela can be run in its own container

liam-lloyd:
For local development, we want to run stela in a consistent environment. This commit enables this by adding a Dockerfile so stela can be run in docker. It also includes hotloading support, so developers won't need to rebuild the docker image in order for changes to take effect.

For the big picture view of how `stela` will integrate with the local env, see: https://github.com/PermanentOrg/devenv/pull/49


cecilia-donnelly (approved):
Thanks for your help debugging this, @liam-lloyd !

PR #9, opened 2023-02-08
PER-9115: Add deployment to dev environment

liam-lloyd:
This commit adds terraform configuration-as-code defining an EKS cluster
in AWS, creating on that cluster a deployment running two pods of stela,
and exposing stela's endpoints to the Internet via an application load
balancer. It also adds workflows to CI to validate the terraform code
and to deploy to dev upon merge to main (or a manual trigger).


terraform/dev/eks-cluster.tf
```diff
@@ -0,0 +1,38 @@
...
+
+  }
+
+  eks_managed_node_groups = {
+    one = {
+      name = "node-group-1"
+
+      vpc_security_group_ids = [var.security_group_id]
+
+      instance_types = ["t3.small"]
```
liam-lloyd, line 31: We might want these instances to be larger when stela starts to have significant workload, but updating them to larger instances should be simple (no worse in impact that our current infrastructure deploys).

terraform/dev/stela_deployment.tf
```diff
@@ -0,0 +1,74 @@
...
+          resources {
+            limits = {
+              cpu    = "1"
+              memory = "512Mi"
+            }
+            requests = {
+              cpu    = "250m"
+              memory = "50Mi"
+            }
+          }
```
liam-lloyd, line 52: Similar to instance size, we'll probably want to increase the resources here in the not-too-distant future

.github/workflows/dev_deploy.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Terraform Plan
+        id: plan
+        if: github.event_name == 'pull_request'
+        run: terraform plan -no-color -input=false
+        continue-on-error: true
+      - name: Terraform Plan Status
+        if: steps.plan.outcome == 'failure'
+        run: exit 1
+      - name: Terraform Apply
+        run: terraform apply -auto-approve -input=false -var="stela_image=$IMAGE_TAG"
```
cecilia-donnelly, line 62: I love having the Terraform plan and apply in here instead of leaving them as manual steps for us. :pray:

terraform/dev/eks-cluster.tf
```diff
@@ -0,0 +1,38 @@
+module "eks" {
+  source  = "terraform-aws-modules/eks/aws"
+  version = "19.0.4"
+
+  cluster_name    = local.cluster_name
+  cluster_version = "1.24"
+
+  vpc_id                         = var.vpc_id
+  subnet_ids                     = ["subnet-a3f202fa", "subnet-fc843999", "subnet-0fc91a78"]
```
cecilia-donnelly, line 9: Where do these come from?

terraform/dev/eks-cluster.tf
```diff
@@ -0,0 +1,38 @@
...
+  cluster_endpoint_public_access = true
+  cluster_security_group_id      = var.security_group_id
+  aws_auth_users = [
+    {
+      userarn  = "arn:aws:iam::364159549467:user/liam"
+      username = "liam"
+      groups   = ["system:masters"]
+    }
```
cecilia-donnelly, line 17: It would be better to use a build-specific user than yours, I think.

terraform/dev/ingress.tf
```diff
@@ -0,0 +1,30 @@
+resource "kubernetes_ingress_v1" "general_ingress" {
+  metadata {
+    name = "general-ingress"
+    annotations = {
+      "alb.ingress.kubernetes.io/scheme"          = "internet-facing"
+      "alb.ingress.kubernetes.io/subnet"          = "subnet-a3f202fa, subnet-fc843999, subnet-0fc91a78"
```
cecilia-donnelly, line 6: Since these are in multiple places, is there a way for us to set them as constants somewhere? I'm imagining that we might want to update them at some point.

terraform/dev/load_balancer.tf
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  set {
+    name  = "vpcId"
+    value = var.vpc_id
+  }
+
+  set {
+    name  = "image.repository"
+    value = "602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon/aws-load-balancer-controller"
```
cecilia-donnelly, line 51: I take it this is the id of the repo you made earlier?

terraform/dev/ingress.tf
cecilia-donnelly: (Oh, maybe in the terraform variables file?)

cecilia-donnelly (commented):
This is so cool! A couple questions/comments. I think the biggest thing is not using your IAM user.

terraform/dev/eks-cluster.tf
liam-lloyd: This isn't actually using my user to do the build, it's telling Terraform that my user should have access to the cluster

terraform/dev/ingress.tf
liam-lloyd: Good idea

terraform/dev/eks-cluster.tf
liam-lloyd: They're properties of our VPC (they basically correspond to different availability zones that our VPC is running in; we're here telling the cluster that it can operate in all these availability zones. We do have a fourth subnet but I didn't use it here because our existing deployments don't use it)

terraform/dev/eks-cluster.tf
cecilia-donnelly: Oooooh interesting, I see! Thanks for the explanation. Is there a way to allow a group of users to have access, then? It seems like it would be good for the whole AWS "Developers" group to have access.

liam-lloyd: Unfortunately, it seems there is not a way to do that.

cecilia-donnelly: So what does it mean for your user to have access to the cluster? Can you add me and Natalie individually if you can't add the group? I don't want to be in a scenario where you're the only one who can access it, but "access" is a broad term so I'm not sure how big a deal it is.

liam-lloyd: Access in this case means "extract information about the cluster from the EKS web interface or `kubectl`". Theoretically it should be pretty rare that we have to do this, but you and Natalie should definitely have access; I'll add you.

liam-lloyd: You should have access now. This should be verifiable by visiting https://us-west-2.console.aws.amazon.com/eks/home?region=us-west-2#/clusters/dev-eks-ZClsj6PM and confirming that there's no message telling you you lack permissions.

cecilia-donnelly: Awesome, thank you! Confirmed.

cecilia-donnelly (approved):
Thanks for setting this up and walking us through it, @liam-lloyd! Very exciting.

PR #11, opened 2023-02-14
Add missing argument to terraform plan step in deploy workflow

liam-lloyd:
This is a fix for a hopefully minor issue with the deploy process that I missed when testing on the feature branch


cecilia-donnelly (approved):
Ah makes sense :crossed_fingers:

PR #12, opened 2023-02-27
Add POST endpoint for creating directives

liam-lloyd:
In pursuit of our goal of adding directives to the Permanent app, this commit adds a POST endpoint for creating a directive (and an associated directive trigger). As the first non-health-check endpoint in this repo, this endpoint establishes a few practices that merit mentioning. The first is using Joi for input validation (we also use Joi in notification-service). Also introduced is a `fixtures` repo, for storing SQL scripts that generate seed data for use in testing. Finally, for ease of development this commit alters tests to create their test database on the `devenv` postgres container, and to copy the schema for that test database from that container's `permanent` database. This is necessary in order to get changes from migrations.

Note: The tests that are failing on this branch will start succeeding once https://github.com/PermanentOrg/devenv/pull/62 is merged


src/directive/service.test.ts
```diff
@@ -0,0 +1,199 @@
...
+  });
+
+  test("should error if type is invalid", async () => {
+    let error = null;
+    try {
+      await directiveService.createDirective({
+        emailFromAuthToken: "test@permanent.org",
+        archiveId: 1,
+        stewardAccountId: 2,
+        type: "delete",
```
cecilia-donnelly, line 99: Is there a reason why you picked "delete" here? I would imagine eventually that won't be an invalid type and this test will have to be updated. Totally fine, of course, just curious.

src/directive/service.ts
```diff
@@ -0,0 +1,92 @@
...
+      } catch (err) {
+        if (isInvalidEnumError(err)) {
+          throw new createError.BadRequest(
+            `${getInvalidValueFromInvalidEnumMessage(
+              err.message
+            )} is not a valid value for "trigger.type"`
+          );
+        }
+        logger.error(err);
+        throw new createError.InternalServerError("Failed to create directive");
```
cecilia-donnelly, line 80: If the trigger fails, should we delete the directive entry?

src/directive/validators.test.ts
```diff
@@ -0,0 +1,233 @@
...
+          type: "admin",
+        },
+      });
+    } catch (err) {
+      error = err;
+    } finally {
+      expect(error).not.toBeNull();
+    }
+  });
+  test("should raise an error if stewardAccountId is missing", () => {
```
cecilia-donnelly, line 37: Is this only for "transfer" type directives?

src/directive/validators.ts
```diff
@@ -0,0 +1,24 @@
+import Joi from "joi";
+import type { CreateDirectiveRequest } from "./model";
+
+export const validateCreateDirectiveRequest = (
+  data: unknown
+): data is CreateDirectiveRequest => {
+  const validation = Joi.object()
+    .keys({
+      archiveId: Joi.number().integer().min(1).required(),
+      stewardAccountId: Joi.number().integer().min(1).required(),
```
cecilia-donnelly, line 10: This answers my test question - this requires the steward's id always, even though we only need it for "transfer" type directives, which of course are the only kind we have right now. I think Joi does handle "if this value then this other value" though. What do you think about using something like that here, so the steward is required if the type is "transfer"? Too complicated for the moment?

src/fixtures/create_test_accounts.sql
```diff
@@ -0,0 +1,5 @@
+INSERT INTO
+  account (accountId, primaryEmail, status, notificationPreferences, type)
+VALUES
+  (2, 'test@permanent.org', 'status.auth.ok', '{}', 'type.account.standard'),  
+  (3, 'test+1@permanent.org', 'status.auth.ok', '{}', 'type.account.standard');
```
cecilia-donnelly, line 5: Any reason not to make these 'type.account.test'?

cecilia-donnelly (commented):
This is so exciting! Leaving some questions since I haven't tried it out yet.

package.json
```diff
@@ -11,10 +11,10 @@
     "start-containers": "docker compose up -d --build stela",
     "clear-database": "psql postgresql://postgres:permanent@localhost:5432 -c 'DROP DATABASE IF EXISTS test_permanent;'",
     "create-database": "psql postgresql://postgres:permanent@localhost:5432 -c 'CREATE DATABASE test_permanent;'",
-    "set-up-database": "psql postgresql://postgres:permanent@localhost:5432/test_permanent -f database/base.sql",
+    "set-up-database": "pg_dump postgresql://postgres:permanent@localhost:5432/permanent --schema public | psql postgresql://postgres:permanent@localhost:5432/test_permanent",
     "clear-database-ci": "psql postgresql://postgres:permanent@database:5432 -c 'DROP DATABASE IF EXISTS test_permanent;'",
     "create-database-ci": "psql postgresql://postgres:permanent@database:5432 -c 'CREATE DATABASE test_permanent;'",
-    "set-up-database-ci": "psql postgresql://postgres:permanent@database:5432/test_permanent -f database/base.sql",
+    "set-up-database-ci": "pg_dump postgresql://postgres:permanent@database:5432/permanent --schema public | psql postgresql://postgres:permanent@database:5432/test_permanent",
```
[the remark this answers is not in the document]
liam-lloyd: Oh, I meant to be using `--schema-only` here, I'll update it to actually do that, which will avoid any local data problems

src/directive/service.ts
liam-lloyd: The `transaction` method is handling that for us, any error thrown in the function it wraps will cause a rollback

src/directive/service.test.ts
liam-lloyd: It was in the spirit of making the test realistic (a planned but not yet supported type seems one of the more likely things to be sent in error). Upon reflection though I think it would be better to change it to something that won't demand updating in the future.

src/directive/validators.ts
liam-lloyd: Good idea, will update

src/fixtures/create_test_accounts.sql
liam-lloyd: I think that as long as we're not actually going to need to filter test accounts out of analysis in a real environment, it's ideal for them to look as much like real accounts as possible.

src/directive/service.ts
cecilia-donnelly: ah, thank you! Sorry I missed that.

package.json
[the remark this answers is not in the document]
cecilia-donnelly: sweet thank you!

meisekimiu (commented):
I haven't gotten tests working locally yet, but leaving a comment to say that I don't see any issues on the Typescript side of things. :+1:

.env.template
```diff
@@ -1,6 +1,6 @@
 PORT=8080
 
-DATABASE_URL="postgres://postgres:permanent@localhost:5432/test_permanent"
+DATABASE_URL="postgres://postgres:permanent@database:5432/test_permanent"
```
cecilia-donnelly, line 3: I think this needs to be `permanent` now, not `test_permanent`, based on our conversations. Thanks again for going through all this with me!

README.md
```diff
@@ -24,19 +24,25 @@ npm install
...
-| FUSIONAUTH_API_KEY                | none                                                        | Can be found in `back-end`'s library/base/constants/base.constants.php |
-| FUSIONAUTH_TENANT                 | none                                                        | Can be found in `back-end`'s library/base/constants/base.constants.php |
-| FUSIONAUTH_BACKEND_APPLICATION_ID | none                                                        | Can be found in `back-end`'s library/base/constants/base.constants.php |
-| FUSIONAUTH_ADMIN_APPLICATION_ID   | none                                                        | Can be found in the FusionAuth Admin application                       |
+| Variable                          | Default                                               | Notes                                                                  |
+| --------------------------------- | ----------------------------------------------------- | ---------------------------------------------------------------------- |
+| DATABASE_URL                      | postgres://postgres:permanent@database:5432/permanent | Run tests to generate default database                                 |
```
cecilia-donnelly, line 29: Oh, I see, you changed it here!

src/directive/validators.ts
```diff
@@ -0,0 +1,28 @@
...
+        otherwise: Joi.valid(null),
+      }),
+      type: Joi.string().required(),
+      note: Joi.string(),
+      trigger: Joi.object()
+        .keys({
+          type: Joi.string().required(),
+        })
+        .required(),
+    })
```
cecilia-donnelly, line 22: So my request failed until I added `emailFromAuthToken: Joi.string().email(),` in the validation rules. Without that, I got:

```
[Error [ValidationError]: "emailFromAuthToken" is not allowed] {
  _original: {
    archiveId: 1,
    stewardAccountId: 3,
    type: 'transfer',
    trigger: { type: 'admin' },
    emailFromAuthToken: 'cecilia+local@permanent.org'
  },
  details: [
    {
      message: '"emailFromAuthToken" is not allowed',
      path: [Array],
      type: 'object.unknown',
      context: [Object]
    }
  ]
}
```

With that new rule, I could send a request without an email and have it succeed, which is what I expected (since I see you're inserting the email from the auth token separately, as only makes sense). I guess Joi limits the body to only the fields in the validator. Presumably you didn't run into this though, so I don't know why it appeared for me!  Maybe there's something else going on?

cecilia-donnelly (changes requested):
I was able to create a directive! Hurray! I did run into this maybe-bug, though - let me know what you think.

src/directive/validators.ts
liam-lloyd: I did run into it, but I apparently fixed it on the wrong branch! Will fix here too.

cecilia-donnelly (approved):
Hurray! Directives are happening!

PR #14, opened 2023-03-08
PER-9159: Back-end: write read endpoint for directives

liam-lloyd:
This is build atop the PER-9148 PR, so if that isn't merged yet compare to `per-9148` to see what's new here.

### DIFF
--- a/src/directive/service/get_by_archive_id.test.ts
+++ b/src/directive/service/get_by_archive_id.test.ts
@@ -1,0 +1,76 @@
+import { NotFound } from "http-errors";
+import { db } from "../../database";
+import { directiveService } from "./index";
+
+jest.mock("../../database");
+
+const testDirectiveId = "39b2a5fa-3508-4030-91b6-21dc6ec7a1ab";
+const testArchiveId = 1;
+const testEmail = "test@permanent.org";
+
+const loadFixtures = async (): Promise<void> => {
+  await db.sql("fixtures.create_test_accounts");
+  await db.sql("fixtures.create_test_archive");
+  await db.sql("fixtures.create_test_account_archive");
+  await db.sql("fixtures.create_test_directive");
+  await db.sql("fixtures.create_test_directive_trigger");
+};
+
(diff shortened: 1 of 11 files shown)


cecilia-donnelly:
I think [6a65abb](https://github.com/PermanentOrg/stela/pull/14/commits/6a65abb083a7e210de6aaf3c4d76dd4ddf7c8637) is the only relevant commit here and the other two are leftover from old versions of the parent branch(es).

cecilia-donnelly (changes requested):
It's possible I've erred in my token management, but I believe that this allows a logged-in user to view the directives associated with *any* archive, which is not what we want. I think a regular user should only be able to see directives associated with archives they own.

(There's maybe a question about an admin being able to see all directives on all archives, but I don't think we've designed that yet.)

src/directive/service/get_by_archive_id.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    );
+    expect(response.length).toBe(1);
+    expect(response[0]?.directiveId).toBe(testDirectiveId);
+  });
+
+  test("should return empty list if archive has no directives", async () => {
+    const response = await directiveService.getDirectivesByArchiveId(9999);
+    expect(response.length).toBe(0);
+  });
+});
```
cecilia-donnelly, line 43: Probably want a "should return empty list if authenticated account does not own archive," assuming that we don't want to confirm a directive exists (i.e., essentially return a 404 rather than a 401)

cecilia-donnelly (commented):
This is missing some fixes from main, I believe because it includes old versions of the "post a directive" and "trigger a directive" commits.

Other than that this behaves as I expect and looks good! I'll officially approve once those other commits are gone/fixed up / whatever needs to happen.

liam-lloyd:
> (Are the package lock changes necessary?)

Oh, I imagine not, since package.json hasn't changed. I'll remove those changes before merging.

PR #15, opened 2023-03-08
Allow tests to access secrets during deploy workflow

liam-lloyd:
It turns out that secrets don't automatically propagate to workflows called from other workflows. This means while the test workflow can check out `back-end` and `devenv` just fine when run on it's own, it can't when called from the deploy workflow. This commit fixes that issue by telling the deploy workflow to let the test workflow see secrets.


cecilia-donnelly (approved):
Nice, thank you!

PR #16, opened 2023-03-08
Add env vars for calls to legacy backend

liam-lloyd:
A previous commit added a client to make calls to `back-end`, but didn't add the necessary env vars for that client to terraform. This commit adds them.

### DIFF
--- a/terraform/dev/variables.tf
+++ b/terraform/dev/variables.tf
@@ -52,11 +52,22 @@ variable "fusionauth_tenant" {
 variable "fusionauth_backend_application_id" {
   description = "ID of the FusionAuth application that manages authentication to the web-app's backend"
   type        = string
-  default     = "0853df7b-d5eb-48d6-bec8-9ec48533b008"
+  default     = "8048057e-4f77-406a-a77d-2962a81cea21"
 }
 
 variable "fusionauth_admin_application_id" {
   description = "ID of the FusionAuth application that manages authentication to the admin portal"
   type        = string
   default     = "f2043bfb-9886-4df8-a0f0-7cd1d75651a0"
+}
+
+variable "legacy_backend_host_url" {
+  description = "Host URL of the legacy PHP backend"
+  type        = string
+  default     = "https://dev.permanent.org/api"
+}
+
+variable "legacy_backend_shared_secret" {
+  description = "Shared secret for authenticating calls to the legacy backend"
+  type        = string
 }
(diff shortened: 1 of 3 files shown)


terraform/dev/variables.tf
```diff
@@ -52,11 +52,22 @@ variable "fusionauth_tenant" {
 variable "fusionauth_backend_application_id" {
   description = "ID of the FusionAuth application that manages authentication to the web-app's backend"
   type        = string
-  default     = "0853df7b-d5eb-48d6-bec8-9ec48533b008"
+  default     = "8048057e-4f77-406a-a77d-2962a81cea21"
```
meisekimiu, line 55: This is just a comment for clarification in case another person reviews this PR as well, since I was confused by this change at first! This is changing the default value for the FusionAuth app ID from local to dev (which makes sense since this config is for dev deployment anyway).

meisekimiu (approved):
This looks good to me, fat least from my understanding of the deployment setup.

terraform/dev/variables.tf
cecilia-donnelly, line 55: Why did this change?

cecilia-donnelly (commented):
This makes sense but I'm not sure how to test it - any advice?

terraform/dev/variables.tf
cecilia-donnelly: (@meisekimiu answered!)

meisekimiu:
@cecilia-donnelly My approval was based on the fact that all the checks passed plus the fact that it looks like @liam-lloyd [ran the Dev Deploy Action on this branch](https://github.com/PermanentOrg/stela/actions/runs/4369942323) and it all worked.

I'm realizing now that there is actually code in the repo right now we can use to test this functionality on: running the "Trigger Admin Directive" endpoint on dev. Do you know how to test that, @cecilia-donnelly?

cecilia-donnelly:
Yeah! Two steps.

1. Create a directive on dev. To create the directive, you POST to https://dev.permanent.org/api/v2/directive using a body like 

```
{
    "stewardAccountId": 3,
    "archiveId": 1,
    "type": "transfer",
    "trigger": {
        "type": "admin"
    }
}
```
and a bearer token that was created using an account on dev. Any account will do, but it should be the owner of the `archiveId`. The `stewardAccountId` is any other account that exists on dev.

2. Trigger the directive by sending a POST request to `https://local.permanent.org/api/v2/directive/trigger/account/{accountId}` where the `accountId` is the account that you used to create the bearer token for step 1. For this step, you need a *different* bearer token that was created with the `admin` client in FusionAuth.

I haven't tried all this on dev yet, just local, but those two steps should do the trick.

PR #17, opened 2023-03-16
Take email not account

cecilia-donnelly:
WIP PR for @meisekimiu. Natalie, this takes the email but doesn't yet return it. I have also broken some linting and tests so it's very in-progress at the moment but thought it might be useful for you anyway. :) 

More: thanks for your patience on this!

### DIFF
--- a/src/directive/service/create.ts
+++ b/src/directive/service/create.ts
@@ -12,6 +12,10 @@ import {
 import { logger } from "../../log";
 import { confirmArchiveOwnership } from "./utils";
 
+interface AccountResult {
+  stewardAccountId: string;
+}
+
 export const createDirective = async (
   requestBody: CreateDirectiveRequest
 ): Promise<Directive> => {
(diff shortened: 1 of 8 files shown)


src/directive/service/create.ts
```diff
@@ -20,6 +24,20 @@ export const createDirective = async (
     requestBody.emailFromAuthToken
   );
 
+  const accountResult = await db.sql<AccountResult>(
+    "directive.queries.check_account_by_email",
+    {
+      email: requestBody.stewardEmail,
+    }
+  );
```
liam-lloyd, line 32: I don't think we should make this query if `type` isn't "transfer". Perhaps even better, we could avoid this query entirely by checking the insert query for an error on the `steward_account_id_not_null_when_transfer` constraint, the same way we handle invalid enum errors.

cecilia-donnelly: That's a good point - thanks for pointing out the invalid enum errors. It should be ready for another look.

liam-lloyd (approved):

PR #18, opened 2023-03-21
Don't try to execute directives that are already successfully executed

liam-lloyd:
Currently, nothing prevents a directive that has been successfully executed from appearing in the list of directives to execute when an account's admin directives are triggered. This would result in a confusing error, as the already executed directive would fail to execute again. To solve this problem, this commit prevents successfully executed directives from being reprocessed.


cecilia-donnelly (approved):
Very helpful, thank you!

PR #19, opened 2023-03-23
Add update endpoint for directives

liam-lloyd:
Directives need to be updateable, so we need an endpoint to update them. This commit adds one. It also updates this repo to handle all IDs as strings, even if they happen to consist of numeric digits. This is for two reasons. First, making the type of a variable `number` should communicate that it represents a mathematical quantity which either represents a real world quantity or will be involved in arithmetic. IDs don't have these qualities, they are just uniquely identifying *strings* of digits. Second, our numeric IDs are bigints in PostgreSQL, and Javascript doesn't natively support integers that large (64 bits). Thus handling IDs as strings avoids cumbersome and potenially error-prone operations to convert PostgreSQL bigints to Javascript numbers.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,10295 +1,8 @@
 {
   "name": "stela",
   "version": "0.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
   "requires": true,
-  "packages": {
-    "": {
-      "name": "stela",
-      "version": "0.0.0",
-      "license": "AGPL-3.0",
-      "dependencies": {
-        "@fusionauth/typescript-client": "^1.42.0",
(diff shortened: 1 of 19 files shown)


liam-lloyd:
> This is nice! I see you only allow updating un-executed directives (hurray!) via the SQL query - what was your thinking about not including an explicit test for that? I ask because I didn't catch that restriction at first glance through the code and it would be more obvious if there was a test, but since it's enforced at the db layer I can see why we wouldn't need one.

I've added a test; I missed it because I was tracing through the paths in the code to determine my test cases and that's enforced in the SQL, but still good to test. In fact, in writing the test I noticed that that case would have come out as an Internal Server Error; I've updated it to be a Bad Request.

PR #22, opened 2023-04-21
Add deployment workflow to deploy stela to staging

liam-lloyd:
In order to make stela a full-fledged component of the Permanent backend, it needs to exist in `staging`. This commit adds a new kubernetes deployment to our dev/staging EKS cluster, with its own ingress (which is bound in AWS to api.staging.permanent.org).


liam-lloyd:
Perhaps worth noting that while we originally planned to separate dev from staging using Kubernetes namespaces, the Kubernetes docs suggested that for an application as simple as ours namespaces wouldn't be the right approach.

.github/workflows/dev_deploy.yml
```diff
@@ -52,8 +52,10 @@ jobs:
       - name: Terraform Validate
         id: validate
         run: terraform validate -no-color
+      # Note that while we do pass a value of stela_staging_image below, it won't be used because the -target option
+      # restricts terraform to just updating the dev deployment
```
cecilia-donnelly, line 56: Interesting, thanks for this. Why include the staging image then?

.github/workflows/staging_deploy.yml
```diff
@@ -0,0 +1,63 @@
...
+        env:
+          AWS_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
+      - name: Terraform Init
+        id: init
+        run: terraform init
+      - name: Terraform Validate
+        id: validate
+        run: terraform validate -no-color
+      # Note that while we do pass a value of stela__image below, it won't be used because the -target option
+      # restricts terraform to just updating the staging deployment
```
cecilia-donnelly, line 58: Missing the word `dev` in this comment I believe.

cecilia-donnelly (approved):
Okay, cool! I see what you mean about this being smaller than you expected. I'm not sure how to test it - I guess make a branch that matches the naming convention for pushing to staging?

Side note: are all of these secrets in the API Key Confluence article you wrote?

.github/workflows/dev_deploy.yml
liam-lloyd: Terraform won't let us leave the variable empty

liam-lloyd:
> Okay, cool! I see what you mean about this being smaller than you expected. I'm not sure how to test it - I guess make a branch that matches the naming convention for pushing to staging?

The way to test would be to disable the restriction of staging deploys to `main` and then create a release that matches the pattern from this branch

> Side note: are all of these secrets in the API Key Confluence article you wrote?

Not all of them, the ones that are purely internal to us (like the DB URI) didn't feel like they were API keys

PR #23, opened 2023-04-26
Add documentation for the directives API

liam-lloyd:


API.md
```diff
@@ -0,0 +1,124 @@
...
+    updatedDt: string (date),
+  },
+  stewardAccountId: string,
+  note: string,
+  executionDt: string (date)
+}
+```
+
+### POST `/directive/trigger/account/:accountId`
+
```
cecilia-donnelly, line 47: This one is a little less intuitive so it might be helpful to add a sentence here like "Execute an existing directive"

cecilia-donnelly: (Thinking more for mobile team than Natalie in this case)

cecilia-donnelly (approved):
Thanks for doing this!

PR #24, opened 2023-04-27
Add create endpoint for legacy contacts

liam-lloyd:
To support our legacy planning feature, users need to be able to record legacy contacts in our system. This commit adds an endpoint for creating such records.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "start-containers": "docker compose up -d --build stela",
     "clear-database": "psql postgresql://postgres:permanent@localhost:5432 -c 'DROP DATABASE IF EXISTS test_permanent;'",
     "create-database": "psql postgresql://postgres:permanent@localhost:5432 -c 'CREATE DATABASE test_permanent;'",
-    "set-up-database": "docker exec --env='PGPASSWORD=permanent' devenv_database_1 pg_dump -U postgres permanent --schema-only | psql postgresql://postgres:permanent@localhost:5432/test_permanent",
+    "set-up-database": "docker exec --env='PGPASSWORD=permanent' devenv-database-1 pg_dump -U postgres permanent --schema-only | psql postgresql://postgres:permanent@localhost:5432/test_permanent",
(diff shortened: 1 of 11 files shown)


liam-lloyd:
The failing tests should start passing when https://github.com/PermanentOrg/back-end/pull/459 is merged

package.json
```diff
@@ -11,7 +11,7 @@
     "start-containers": "docker compose up -d --build stela",
     "clear-database": "psql postgresql://postgres:permanent@localhost:5432 -c 'DROP DATABASE IF EXISTS test_permanent;'",
     "create-database": "psql postgresql://postgres:permanent@localhost:5432 -c 'CREATE DATABASE test_permanent;'",
-    "set-up-database": "docker exec --env='PGPASSWORD=permanent' devenv_database_1 pg_dump -U postgres permanent --schema-only | psql postgresql://postgres:permanent@localhost:5432/test_permanent",
+    "set-up-database": "docker exec --env='PGPASSWORD=permanent' devenv-database-1 pg_dump -U postgres permanent --schema-only | psql postgresql://postgres:permanent@localhost:5432/test_permanent",
```
liam-lloyd, line 14: Not sure how this ended up wrong, but container names use dashes, not underscores

src/legacy_contact/service/create.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    try {
+      await legacyContactService.createLegacyContact({
+        emailFromAuthToken: "not_an_account@permanent.org",
+        name: "Legacy Contact",
+        email: "legacy.contact@permanent.org",
+      });
+    } catch (err) {
+      error = err;
+    } finally {
+      expect(error instanceof InternalServerError).toBe(true);
```
liam-lloyd, line 60: I can see cases for using a different type of error if the email from the auth token doesn't have an account. My reasoning for using `InternalServerError` is that this is an error we wouldn't expect to happen (no one should be able to get a valid JWT without having an account), and if it does it suggests something is wrong with our JWTs, which seems like a risky situation in which we should reveal as little as possible.

meisekimiu (approved):

PR #25, opened 2023-05-05
Add PUT /account/tags endpoint to update an account's Mailchimp tags

liam-lloyd:
When users set goals in a future version of the onboarding flow, we want to record those goals as tags in Mailchimp. As such, we need an endpoint allowing a logged in user to add tags to themself in Mailchimp. This commit adds such an endpoint.


src/types/mailchimp-marketing.d.ts
```diff
@@ -0,0 +1,23 @@
+declare module "@mailchimp/mailchimp_marketing" {
```
liam-lloyd, line 1: There is a published type file for this module, but it has the wrong type for `updateListMemberTags`

src/account/validators.test.ts
```diff
@@ -0,0 +1,115 @@
...
+      error = err;
+    } finally {
+      expect(error).not.toBeNull();
+    }
+  });
+  test("should raise an error if addTags is not an array", () => {
+    let error = null;
+    try {
+      validateUpdateTagsRequest({
+        emailFromAuthToken: "not_an_email",
```
cecilia-donnelly, line 63: Copy/paste error here and below, I think - these should have valid emails so we're testing what we claim to.

src/account/validators.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const validateUpdateTagsRequest = (
+  data: unknown
+): data is UpdateTagsRequest => {
+  const validation = Joi.object()
+    .keys({
+      emailFromAuthToken: Joi.string().email().required(),
+      addTags: Joi.array().items(Joi.string()),
+      removeTags: Joi.array().items(Joi.string()),
+    })
```
cecilia-donnelly, line 12: Should we validate that we either add or remove tags on each request? I guess it doesn't really matter, but it would be weird to call this endpoint with no tags at all.

src/types/mailchimp-marketing.d.ts
cecilia-donnelly: Is there an upstream where you could open a PR, if that's fast?

terraform/test_cluster/stela_dev_deployment.tf
```diff
@@ -83,6 +83,27 @@ resource "kubernetes_deployment" "stela_dev" {
...
+          env {
+            name  = "MAILCHIMP_DATACENTER"
+            value = var.mailchimp_datacenter
+          }
+
+          env {
+            name  = "MAILCHIMP_COMMUNITY_LIST_ID"
+            value = var.dev_mailchimp_community_list_id
+          }
+
```
cecilia-donnelly, line 106: This is interesting - so each time we add an environment variable to stela we should also add it in these deployment files?

cecilia-donnelly (commented):
A few comments before I test.

terraform/test_cluster/stela_dev_deployment.tf
liam-lloyd: Yes

src/types/mailchimp-marketing.d.ts
liam-lloyd: There is, but it might require investigating several other endpoints to do properly, because I suspect there are others with the same problem.

cecilia-donnelly (commented):
My default is to add new environment variables to .env.template, to help someone setting up the project for the first time. Can you do that, or do you think that's putting the same information in too many places?

I also got a suggestion from typescript to run `npm i --save-dev @types/mailchimp__mailchimp_marketing` to add the types to package.json so it could find them, which was easy enough, but that line in package.json should be in this PR.

And I get this error:
```
src/account/service.ts(28,32): error TS2339: Property 'status' does not exist on type '{} | ErrorResponse'.
  Property 'status' does not exist on type '{}'.
```

And the same thing for `detail`. It's a bit weird since you're just pulling in from `http-errors`, so I don't know why it's complaining. I tried rebasing on main and that unfortunately did not fix it. Does that look familiar?

liam-lloyd:
> My default is to add new environment variables to .env.template, to help someone setting up the project for the first time. Can you do that, or do you think that's putting the same information in too many places?
> 
> I also got a suggestion from typescript to run `npm i --save-dev @types/mailchimp__mailchimp_marketing` to add the types to package.json so it could find them, which was easy enough, but that line in package.json should be in this PR.
> 
> And I get this error:
> 
> ```
> src/account/service.ts(28,32): error TS2339: Property 'status' does not exist on type '{} | ErrorResponse'.
>   Property 'status' does not exist on type '{}'.
> ```
> 
> And the same thing for `detail`. It's a bit weird since you're just pulling in from `http-errors`, so I don't know why it's complaining. I tried rebasing on main and that unfortunately did not fix it. Does that look familiar?

Added new env vars to the `.env.template` and also the README. Both of the type related errors where caused because we needed an extra setting in `tsconfig.json` to ensure `ts-node` can always find our custom type files, this should be fixed now.

cecilia-donnelly (commented):
I thought the endpoint would be https://local.permanent.org/api/v2/account/2/tags for an accountid of 2, but apparently not since I get a 404 - where is it? (I know it's wild to ask this on the third go-round of review, sorry!)

liam-lloyd:
> I thought the endpoint would be https://local.permanent.org/api/v2/account/2/tags for an accountid of 2, but apparently not since I get a 404 - where is it? (I know it's wild to ask this on the third go-round of review, sorry!)

It'll just be https://local.permanent.org/api/v2/account/tags; there's no need to specify the account id in the URL because the account is identified by the auth token.

src/account/service.ts
```diff
@@ -0,0 +1,34 @@
...
+          status: "inactive",
+        })
+      )
+    );
+
+  const response = await MailChimpClient.lists.updateListMemberTags(
+    process.env["MAILCHIMP_COMMUNITY_LIST_ID"] ?? "",
+    Md5.hashStr(requestBody.emailFromAuthToken),
+    { tags }
+  );
```
cecilia-donnelly, line 25: I think this should be in a try/catch or otherwise handle a MailChimp failure a little more intentionally. My email wasn't present on Mailchimp (because it's a local test email, I suppose, and we don't generally add those to Mailchimp) and so I got a 404 back. If something goes wrong here then I think it would be good to catch that and send a JSON error. Currently we just pass on the "not found" HTML.

cecilia-donnelly (commented):
Quick thought.

src/account/service.ts
liam-lloyd: Added middleware to convert these and any other third party errors with `status` or `statusCode` fields to JSON

cecilia-donnelly (approved):
Success! Got this to add a tag to a contact in MailChimp. Phew.

PR #26, opened 2023-05-08
Fix separation between staging and dev in the test EKS cluster

liam-lloyd:
Currently, requests to `stela` endpoints in `dev` or `staging` can end up routed to a pod from the other environment. This commit should fix that.


cecilia-donnelly (approved):
This makes sense. Thank you for investigating, @liam-lloyd !

PR #27, opened 2023-05-09
Return errors as JSON

liam-lloyd:
Since stela endpoints return their normal responses and their validation errors as JSON, they should return other errors as JSON too so clients only need to parse one format when consuming this API. This commit adds middleware to translate HTTP errors into JSON before returning them to the client. It also adds documentation on our error formats to the API documentation file.

### DIFF
--- a/API.md
+++ b/API.md
@@ -2,6 +2,18 @@
 
 - Path: `<host>/api/v2`
 
+## Errors
+Errors that can be foreseen, including any errors intended to be part of client control flow, will come with a JSON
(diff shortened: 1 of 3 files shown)


cecilia-donnelly:
@luciancerbu-vsp this should get merged today

cecilia-donnelly (approved):
Thank you! I'll merge since I told @luciancerbu-vsp I would, though I realize now that this isn't on a specially-named branch so probably won't be deployed automatically.

PR #28, opened 2023-05-10
Add endpoint to get legacy contact(s) by accountId

liam-lloyd:
As we build out the legacy contact UI, clients will need to retrieve an account's legacy contact and display it to the user. This commit adds an endpoint to enable that.


cecilia-donnelly (changes requested):
This needs an entry in API.md to explain how to call it. Would you also add one for creating a legacy contact, since I missed it in that PR?

We only expect an account to have one legacy contact, right? Why return an array? (I'm sure this is because we talked about it and decided to do it that way, but remind me!)

cecilia-donnelly:
Oh, and other than that this looks good! I was able to call it and get the expected legacy contact. :)

liam-lloyd:
> We only expect an account to have one legacy contact, right? Why return an array? (I'm sure this is because we talked about it and decided to do it that way, but remind me!)

My thinking was that, while current designs expect no more than one legacy contact, it seems pretty likely to me that we'll want to give people the option of having more than one in the future.

cecilia-donnelly (approved):
Thanks!

PR #29, opened 2023-05-10
PER-9204: Set up update endpoint for legacy contacts

liam-lloyd:
This is built atop the per-9203 branch; the parts unique to this PR are in the second commit.


cecilia-donnelly (changes requested):
Looks good, though also needs an entry in API.md. Thanks @liam-lloyd !

cecilia-donnelly (approved):
And double thanks!

PR #30, opened 2023-05-12
Return steward name and email on GET

cecilia-donnelly:
Also update documentation. The other endpoints were correctly returning the email instead of the account id, but this one was not. @luciancerbu-vsp also requested that we return the name, so I included that - let me know if that'll be a problem.

### DIFF
--- a/API.md
+++ b/API.md
@@ -49,7 +49,7 @@ instead return a [Joi ValidationError object](https://joi.dev/api/?v=17.9.1#vali
     createdDt: string (date),
     updatedDt: string (date),
   },
-  stewardAccountId: string,
+  stewardEmail: string,
   note: string,
   executionDt: string (date)
 }
@@ -104,7 +104,7 @@ instead return a [Joi ValidationError object](https://joi.dev/api/?v=17.9.1#vali
     createdDt: string (date),
     updatedDt: string (date),
   },
-  stewardAccountId: string,
+  stewardEmail: string,
   note: string,
   executionDt: string (date)
 }
@@ -130,7 +130,8 @@ instead return a [Joi ValidationError object](https://joi.dev/api/?v=17.9.1#vali
       createdDt: string (date),
       updatedDt: string (date),
     },
-    stewardAccountId: string,
+    stewardEmail: string,
+    stewardName: string,
     note: string,
     executionDt: string (date)
   }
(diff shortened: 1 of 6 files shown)


API.md
```diff
@@ -129,7 +129,8 @@ instead return a [Joi ValidationError object](https://joi.dev/api/?v=17.9.1#vali
       createdDt: string (date),
       updatedDt: string (date),
     },
-    stewardAccountId: string,
+    stewardEmail: string,
+    stewardName: string,
```
liam-lloyd, line 133: If we're returning multiple properties of the steward, I think we should make the steward its own object within our directive object.
```
steward: {
  email: string,
  name: string,
}
```

liam-lloyd: Of course that probably breaks `web-app`'s integration with this endpoint, which isn't ideal.

cecilia-donnelly: That makes sense - @meisekimiu let's discuss in planning.

cecilia-donnelly:
Updated to return a steward object on GET. I didn't change the other endpoints to include the name, though maybe I should. What do you think, @luciancerbu-vsp and @meisekimiu ?

luciancerbu-vsp:
> Updated to return a steward object on GET. I didn't change the other endpoints to include the name, though maybe I should. What do you think, @luciancerbu-vsp and @meisekimiu ?

I think it will be nice for the other endpoints also to include the name.

cecilia-donnelly:
Updated so creating, updating, and getting a directive all return a steward object with name and email.

liam-lloyd (commented):
Everything looks good, but we should update the `Directive` interface in `src/directive/model.ts` to reflect this change - unfortunately because Typescript doesn't enforce itself at runtime nothing actually breaks if the type we say we're returning from these endpoints isn't what we actually get back from the database.

cecilia-donnelly:
Added that type change - thanks for pointing it out! Didn't even think of it.

src/directive/model.ts
```diff
@@ -35,6 +35,7 @@ export interface Directive {
   updatedDt: Date;
   trigger: DirectiveTrigger;
   stewardAccountId?: string;
```
liam-lloyd, line 37: Commented on the commit already (unintentionally, I didn't know you could do that until just now), but I think we should remove the `stewardAccountId` if it's only used in tests; tests shouldn't be able to influence what our models look like.

cecilia-donnelly: Done.

liam-lloyd (approved):

PR #32, opened 2023-05-23
Add prod deployments

liam-lloyd:
We want to use `stela` for real, so it needs to be deployed to prod. This commit sets up the prod deploy workflow. It also adds an "all test environments" deploy workflow, to handle cases where we need to update parts of the test cluster that our targeted deploys to dev and staging won't affect.

### DIFF
--- a/.github/workflows/prod_deploy.yml
+++ b/.github/workflows/prod_deploy.yml
@@ -1,6 +1,8 @@
-name: Deploy to staging
+name: Deploy to prod
 on:
   workflow_dispatch:
+  release:
+    types: [published]
 jobs:
   run_tests:
     uses: ./.github/workflows/test.yml
@@ -26,7 +28,7 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
       - name: Publish Image to ECR
         run: docker push $IMAGE_TAG
-  deploy:
+  deploy_staging:
     needs:
       - build
     runs-on: ubuntu-20.04
@@ -56,3 +58,33 @@ jobs:
         run: terraform plan -no-color -input=false -var="stela_dev_image=$IMAGE_TAG" -var="stela_staging_image=$IMAGE_TAG" -target=kubernetes_deployment.stela_staging
       - name: Terraform Apply
         run: terraform apply -auto-approve -input=false -var="stela_dev_image=$IMAGE_TAG" -var="stela_staging_image=$IMAGE_TAG" -target=kubernetes_deployment.stela_staging
+  deploy_prod:
+    needs:
+      - build
+      - deploy_staging
+    runs-on: ubuntu-20.04
+    environment: prod
+    defaults:
+      run:
+        working-directory: ./terraform/prod_cluster
+    steps:
+      - uses: actions/checkout@v3
+      - name: Setup Terraform
+        uses: hashicorp/setup-terraform@v1
+        with:
+          cli_config_credentials_token: ${{ secrets.TERRAFORM_API_TOKEN }}
+      - name: Generate Image Tag
+        run: echo "IMAGE_TAG=364159549467.dkr.ecr.$AWS_REGION.amazonaws.com/stela:$([[ ${GITHUB_REF##*/} = main ]] && echo main || echo feature)-$(git rev-parse --short $GITHUB_SHA)" >> $GITHUB_ENV
+        env:
+          AWS_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
+      - name: Terraform Init
+        id: init
+        run: terraform init
+      - name: Terraform Validate
+        id: validate
+        run: terraform validate -no-color
+      - name: Terraform Plan
+        id: plan
+        run: terraform plan -no-color -input=false -var="stela_image=$IMAGE_TAG"
+      - name: Terraform Apply
+        run: terraform apply -auto-approve -input=false -var="stela_image=$IMAGE_TAG"
--- a/README.md
+++ b/README.md
@@ -71,3 +71,22 @@ Outside a container: Run
 ```bash
 npm run start
 ```
+
+## Deployment
+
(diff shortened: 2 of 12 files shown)


.github/workflows/staging_deploy.yml
```diff
@@ -1,11 +1,6 @@
 name: Deploy to staging
 on:
   workflow_dispatch:
-  release:
-    branches:
-      - main
-    tags:
-      - v[0-9].[0-9]+.[0-9]+-staging
```
liam-lloyd, line 8: Turns out the `release` trigger doesn't actually support filters, so there's no way to have this workflow run on any kind of `release` without also triggering the prod workflow. Since deploys to `staging` should mostly happen as part of the `prod` deploy workflow anyway, I've made this workflow manually triggered only.

.github/workflows/prod_deploy.yml
```diff
@@ -0,0 +1,90 @@
...
+    secrets: inherit
+
+  build:
+    needs:
+      - run_tests
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+      - name: Generate Image Tag
+        run: echo "IMAGE_TAG=364159549467.dkr.ecr.$AWS_REGION.amazonaws.com/stela:$([[ ${GITHUB_REF##*/} = main ]] && echo main || echo feature)-$(git rev-parse --short $GITHUB_SHA)" >> $GITHUB_ENV
```
cecilia-donnelly, line 18: So we'll be able to publish a feature branch to prod? (That seems plausible in case of a hotfix but I'm mainly asking in case this was a copy paste error!)

.github/workflows/prod_deploy.yml
```diff
@@ -0,0 +1,90 @@
...
+        run: docker build -t $IMAGE_TAG .
+      - name: AWS Login
+        run: aws ecr get-login-password --region $AWS_REGION | docker login --username AWS --password-stdin 364159549467.dkr.ecr.$AWS_REGION.amazonaws.com
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      - name: Publish Image to ECR
+        run: docker push $IMAGE_TAG
+  deploy_staging:
```
cecilia-donnelly, line 31: Why do staging as part of the prod deploy?

.github/workflows/staging_deploy.yml
cecilia-donnelly: Interesting!

README.md
```diff
@@ -71,3 +71,22 @@ Outside a container: Run
...
+
+## Deployment
+
+### To `dev`
+`stela` deploys to `dev` automatically upon any merge to `main`.
+
+### Beyond `dev`
+To deploy to `staging` and `prod`, create a Release with a tag of the form `vX.X.X` (this should conform to semantic
+versioning). This will trigger a deploy to staging. If that is successful, the workflow will pause and wait for manual
+approval to deploy to prod. Given manual approval, it will deploy to prod.
```
cecilia-donnelly, line 83: So this should be updated to say that releases don't automatically trigger a deploy to staging? Or am I misunderstanding your comment above? I guess I've *seen* releases automatically trigger a deploy to staging, so I know that works.

README.md
```diff
@@ -71,3 +71,22 @@ Outside a container: Run
...
+
+### To `staging` Only
+To deploy to `staging` without deploying to prod, run the "Deploy to staging" workflow from the Actions tab.
+
+### Updating Test Cluster Infrastructure
+Because `dev` and `staging` run on the same EKS cluster, deploys to each of these environments just target the `stela`
+deployments and won't update the underlying infrastructure they run on. To update that infrastructure, manually trigger
+the "Deploy to all test envs" workflow from the Actions tab in Github. This will deploy the current `main` branch to
+`dev` and `staging` in addition to updating the test cluster infrastructure.
```
cecilia-donnelly, line 92: Awesome, thank you for documenting this!

terraform/prod_cluster/eks-cluster.tf
```diff
@@ -0,0 +1,48 @@
...
+      userarn  = "arn:aws:iam::364159549467:user/cecilia"
+      username = "cecilia"
+      groups   = ["system:masters"]
+    },
+    {
+      userarn  = "arn:aws:iam::364159549467:user/natalie"
+      username = "natalie"
+      groups   = ["system:masters"]
+    }
+  ]
```
cecilia-donnelly, line 28: Note that we should add removing us from this list to the Offboarding Jira epic - not urgent since the person's Amazon account would be shut down on departure, but a nice cleanliness thing.

.github/workflows/prod_deploy.yml
cecilia-donnelly: (never mind, you answered this later!)

.github/workflows/prod_deploy.yml
cecilia-donnelly: I think this is because we might want to send feature branches to *staging*, now that I've read the whole PR, but let me know if that's right.

cecilia-donnelly (approved):
Some questions inline but this makes sense to me! Thanks @liam-lloyd .

README.md
liam-lloyd: Releases still trigger a deploy to staging, there just isn't a way to do deploys that are _only_ going to staging through releases

.github/workflows/prod_deploy.yml
liam-lloyd: This shouldn't in practice run on non-`main` branches (and there are repo-level constraints in place to block the prod deploy if it is), but if it does we want to be sure that a build from a feature branch wouldn't get tagged with "main"

PR #33, opened 2023-06-01
Send note and isLegacyAction flag to /archive/transferOwnership

liam-lloyd:
We are updating the /archive/transferOwnership endpoint to send a separate email when the transfer is happening as part of a legacy plan. This commit updates the call to /archive/transferOwnership upon executing a directive to identify the action as part of a legacy call, and includes the note associated with the directive so it can be included in the email.


liam-lloyd:
Relies on https://github.com/PermanentOrg/back-end/pull/473

cecilia-donnelly (approved):
LGTM!

PR #35, opened 2023-06-06
Add CORS headers to stela endpoints

liam-lloyd:


cecilia-donnelly (changes requested):
I know we're adding the environment variables a few places at this point, but can you add ENV to .env.template, too? Without it in .env this will crash locally.

cecilia-donnelly (approved):
Thanks!

PR #36, opened 2023-06-08
QA: Allow removing note to steward

cecilia-donnelly:
@k8lyn6 found in QA that removing a note for a steward throws an error. Joi does not consider the empty string to be a string unless explicitly told to do so. With this change, users can clear a previously-written note.


liam-lloyd (approved):
Prettier wants double quotes but otherwise looks good

cecilia-donnelly:
Oops, thanks!

PR #37, opened 2023-06-12
Send notification email to new legacy contacts

liam-lloyd:
We want to notify people when they become a Permanent user's legacy contact. This commit updates the create endpoint for legacy contacts to send a notification email to the new legacy contact, and updates the update endpoint for legacy contacts to send the same notification email if the update changed the email of the legacy contact.


liam-lloyd:
Needs https://bitbucket.org/permanent-org/email/pull-requests/50 before it will actually send an email

cecilia-donnelly (commented):
This looks good for the most part! I see that the notification email comes through with no subject line, though (at least for me locally). I know @liam-lloyd and @k8lyn6 were trying to figure out how to set the subject line with our various services. Is this likely to be a local problem or is it a blocker? (I guess, does it look right to you, @liam-lloyd ?)

liam-lloyd:
> I see that the notification email comes through with no subject line

Are you sure you're on the latest version of this branch? I encountered that problem this morning and pushed a fix that should have resolved it.

cecilia-donnelly (approved):
Sorry I missed that last subject line fix! Thanks for pointing it out, @liam-lloyd .

PR #38, opened 2023-06-13
Return error when directive update uses email with no account

liam-lloyd:
The steward email associated with a directive is required to belong to a Permanent account. If a steward email without an account is passed to the directive update endpoint, it will presently return success but decline to update the email. This is not very clear to the caller. To improve clarity, this commit updates that endpoint to return an invalid request error in this case.


cecilia-donnelly (approved):
Thank you!

PR #39, opened 2023-06-13
Prevent accounts from being their own legacy contact or archive steward

liam-lloyd:
It doesn't make sense for a user to be their own legacy contact or the steward of their own archive, as these roles are intended to take actions after the user has passed away or become incapacitated. Thus, this commit adds a validation check to prevent users from becoming their own legacy contact the steward of their own archive.


cecilia-donnelly (approved):
Nice!

PR #40, opened 2023-06-13
Add Mailchimp Transactional API key to terraform secrets

liam-lloyd:
Now that we're using Mailchimp Transactional in this project, terraform needs to know about its API key so it can supply it to the deployments. This commit adds the API key to terraform's secrets.

### DIFF
--- a/terraform/test_cluster/secrets.tf
+++ b/terraform/test_cluster/secrets.tf
@@ -4,10 +4,11 @@ resource "kubernetes_secret" "dev-secrets" {
   }
 
   data = {
-    "DATABASE_URL"                 = var.dev_database_url
-    "FUSIONAUTH_API_KEY"           = var.dev_fusionauth_api_key
-    "LEGACY_BACKEND_SHARED_SECRET" = var.dev_legacy_backend_shared_secret
-    "MAILCHIMP_API_KEY"            = var.mailchimp_api_key
+    "DATABASE_URL"                    = var.dev_database_url
+    "FUSIONAUTH_API_KEY"              = var.dev_fusionauth_api_key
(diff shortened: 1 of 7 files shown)


terraform/test_cluster/secrets.tf
```diff
@@ -17,9 +18,9 @@ resource "kubernetes_secret" "staging-secrets" {
...
 
   data = {
-    "DATABASE_URL"                 = var.staging_database_url
-    "FUSIONAUTH_API_KEY"           = var.staging_fusionauth_api_key
-    "LEGACY_BACKEND_SHARED_SECRET" = var.staging_legacy_backend_shared_secret
-    "MAILCHIMP_API_KEY"            = var.mailchimp_api_key
+    "DATABASE_URL"                    = var.staging_database_url
+    "FUSIONAUTH_API_KEY"              = var.staging_fusionauth_api_key
+    "LEGACY_BACKEND_SHARED_SECRET"    = var.staging_legacy_backend_shared_secret
+    "MAILCHIMP_TRANSACTIONAL_API_KEY" = var.mailchimp_transactional_api_key
```
cecilia-donnelly, line 24: Elsewhere we have both the mailchimp API key and the transactional API key - should we have both here too?

cecilia-donnelly (commented):
Thank you! Sorry I didn't think of this. One question but generally makes sense.

terraform/test_cluster/secrets.tf
liam-lloyd: Yes, we should, didn't realize the marketing key wasn't there yet

liam-lloyd: Oh, it was there and I accidentally replaced it. Good catch!

cecilia-donnelly (approved):

PR #41, opened 2023-06-14
Send notification emails to archive stewards

liam-lloyd:
We want to send notification emails to users when they become archive stewards. This commit updates the create and update endpoints of directives to send such emails.


cecilia-donnelly (approved):
Exciting!

PR #42, opened 2023-06-14
Allow www origin for production requests

cecilia-donnelly:
We redirect permanent.org to www.permanent.org, so should expect the latter to be the origin for requests.

Reviewers, should we also allow requests from plain permanent.org?

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -11,7 +11,7 @@ const env = process.env["ENV"] ?? "";
 const app = express();
 app.use(
   cors({
-    origin: `https://${env === "production" ? "" : `${env}.`}permanent.org`,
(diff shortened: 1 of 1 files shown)


src/app.ts
```diff
@@ -11,7 +11,7 @@ const env = process.env["ENV"] ?? "";
 const app = express();
 app.use(
   cors({
-    origin: `https://${env === "production" ? "" : `${env}.`}permanent.org`,
+    origin: `https://${env === "production" ? "www" : `${env}.`}permanent.org`,
```
liam-lloyd, line 14: I think we need a `.` prefixing the `permanent.org` now (in place of the `.` after `${env}`

cecilia-donnelly: oh thanks!

liam-lloyd (approved):

PR #43, opened 2023-06-21
PER-9225: Create endpoint to check which promo code the user signed up with

liam-lloyd:


API.md
```diff
@@ -215,3 +215,35 @@ instead return a [Joi ValidationError object](https://joi.dev/api/?v=17.9.1#vali
...
+}
+```
+
+- Response
+
+```
+{} (empty response with 200 code indicates success)
+```
+
+### GET `/signups`
```
cecilia-donnelly, line 239: Looks like this is singular?

cecilia-donnelly: I also think both of these would be clearer if they were written as `account/tags` and `account/signup`, as we do with `/directive/archive/:archiveId` above, even though it's in the "Directives" section of the docs.

cecilia-donnelly (approved):
Thank you for reorganizing validation! Aside from the documentation comments this makes a ton of sense. I was able to get the signup code for a user I created locally via an invitation. Exciting! I'll mark this as an approval to reduce overhead since I don't want it to sit while I'm gone for an extra "s."

PR #44, opened 2023-06-22
Add GET /archive/:archiveId/tags/public endpoint

liam-lloyd:
To facilitate searching public archives by tag, clients need to be able to know what tags exist in a public archive. This commit adds an endpoint that provides that information.


cecilia-donnelly (approved):
This all looks good! I broke my local (space management problems, not to do with this branch) and want to record my +1 anyway, but for full transparency I have not run this.

PR #45, opened 2023-06-30
Add GET /archive/:archiveId/payer-account-storage endpoint

liam-lloyd:
For archives with payer accounts, we'd like to display the remaining storage available to that archive. To that end, this commit adds an endpoint that returns the storage available to the payer account of an archive, if that archive has a payer account.

### DIFF
--- a/API.md
+++ b/API.md
@@ -253,6 +253,7 @@ instead return a [Joi ValidationError object](https://joi.dev/api/?v=17.9.1#vali
 ### GET `/archive/:archiveId/tags/public`
 
 - Response
+
 ```
 [
   {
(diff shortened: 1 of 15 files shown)


liam-lloyd:
Tests currently fail due to reliance on https://github.com/PermanentOrg/back-end/pull/475

API.md
```diff
@@ -268,3 +269,27 @@ instead return a [Joi ValidationError object](https://joi.dev/api/?v=17.9.1#vali
...
+- Response
+
+```
+{
+  accountSpaceId: string,
+  accountId: string,
+  spaceLeft: string,
+  spaceTotal: string,
+  filesLeft: string,
+  filesTotal: string,
```
liam-lloyd, line 286: I can see an argument that these should be numbers instead, as they do represent a real thing that is numeric. My reasoning for leaving them strings is that this endpoint isn't doing any math on them, it's just retrieving them for display. Further, Javascript numbers are floating point by default, so reliably maintaining their precision involves some extra overhead.

liam-lloyd:
> Tests currently fail due to reliance on [PermanentOrg/back-end#475](https://github.com/PermanentOrg/back-end/pull/475)

Tests passing now

cecilia-donnelly (approved):
Thank you!

PR #47, opened 2023-07-11
Add Sentry

liam-lloyd:
stela should log its errors to Sentry, like our other projects do. This commit adds that logging to stela's middleware.


liam-lloyd:
Note that presently we are (or at least I am) getting email alerts for errors in all `stela` envs; once this is deployed to prod it will create the production environment in the `stela` Sentry project, and we'll then be able to restrict alerts to that environment

PR #48, opened 2023-07-17
Revert "Add GET /archive/:archiveId/payer-account-storage endpoint"

liam-lloyd:
Reverts PermanentOrg/stela#45


liam-lloyd:
The PR this reverts wasn't actually ready to merge because we backed out the `back-end` PR that it depended on

PR #51, opened 2023-07-24
Don't run terraform validation if terraform hasn't changed

liam-lloyd:


liam-lloyd:
Aside from just not doing things that don't need done, the point of this is to make dependabot PRs work, because they can't access terraform to run terraform validation, but also reliably won't change terraform

cecilia-donnelly (approved):
Nice, thanks!

PR #57, opened 2023-10-28
Add notification email send for gift recipients

liam-lloyd:
When a user receives a gift from an other user, we want to send them an email notification. This commit updates /billing/gift to send those email notifications to gift recipients.


cecilia-donnelly (approved):
This looks good! Thanks @liam-lloyd .

PR #58, opened 2023-11-15
Fix test coverage analysis

liam-lloyd:
When codecov was installed in this repo, the tests weren't yet running in their own docker container. When the tests were moved to a container, the coverage report wasn't sent from that container back to the Github actions box where the codecov action runs, so coverage reports stopped reaching codecov. This commit fixes this by voluming the coverage directory on the test container.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,5 +33,5 @@ jobs:
             - run: touch stela/.env
             - run: (cd devenv; docker compose up database_setup -d; docker logs devenv-database_setup-1)
             - run: (cd stela; npm run start-containers)
-            - run: (cd stela; docker compose run stela npm run test-ci --coverage)
+            - run: (cd stela; docker compose run stela npm run test-ci)
             - uses: codecov/codecov-action@v2
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,6 +8,8 @@ services:
       - devenv
     environment:
       - ENV=test
+    volumes:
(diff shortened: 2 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -33,5 +33,5 @@ jobs:
             - run: touch stela/.env
             - run: (cd devenv; docker compose up database_setup -d; docker logs devenv-database_setup-1)
             - run: (cd stela; npm run start-containers)
-            - run: (cd stela; docker compose run stela npm run test-ci --coverage)
+            - run: (cd stela; docker compose run stela npm run test-ci)
```
liam-lloyd, line 36: --coverage is already passed in the test-ci script, so unnecessary here.

liam-lloyd:
Planning to ignore the check that test coverage didn't decline here; we don't want to lock merging this behind reaching 100% test coverage

meisekimiu:
Should unit tests also run when merging main? I think I had to do that on the web-app to make sure that codecov was always comparing to the most up to date version of main.

liam-lloyd:
> Should unit tests also run when merging main? I think I had to do that on the web-app to make sure that codecov was always comparing to the most up to date version of main.

They do run on merge to main (the test workflow is invoked from the dev deploy workflow; it doesn't trigger independently on main to prevent it from running twice)

cecilia-donnelly (approved):
Thanks @liam-lloyd !

PR #59, opened 2023-11-17
Update indentation so these run

cecilia-donnelly:
I noticed that the Terraform validation didn't run on a couple recent PRs (e.g. [here](https://github.com/PermanentOrg/stela/actions/runs/6673486682)) and I think this indentation change might fix that.


liam-lloyd (approved):
Good catch! I was wondering why we got errors about this workflow in cases where it wasn't supposed to run (that is, where the terraform wasn't updated)

PR #61, opened 2023-12-06
Update archive.public to a boolean in base.sql

liam-lloyd:
In base.sql, the file from which we build our local database, the column archive.public has the type integer, while in deployed environments it is a boolean. This commit corrects that error.


liam-lloyd:
Note that I did look for any other errors of the same class, where 1-bit integers from the old MySQL database because booleans in deployed environments but not locally, and did not find any

meisekimiu (approved):

PR #62, opened 2023-12-20
Broke the api server and the logger into separate packages

liam-lloyd:
As we add lambdas to stela, we'll want to be able to use some code, like the logger, in both the api server and the lambdas. This commit separates the logger from the api server to allow for that.


liam-lloyd:
I'm happy to do a walkthrough of this; it's not that huge of a change by total volume, but a lot of it is using techniques that were new to me.

liam-lloyd:
I recommend reviewing the commits separately here, as they are essentially unrelated. The second and third are just here to make test coverage go up to appease codecov.

meisekimiu (approved):
This seems to be working for me; I can run the API server and lint and run tests for both workspaces on my local machine just fine.

PR #63, opened 2024-01-09
Update EKS test cluster to 1.25

liam-lloyd:
Kubernetes v1.24 is about to fall out of standard support in EKS, so we need to upgrade. This commit upgrades just the test cluster, so we can ensure the upgrade doesn't cause problems before upgrading on production.


cecilia-donnelly (approved):
Makes sense to me!

PR #65, opened 2024-01-11
Add linting for SQL files

liam-lloyd:
In the interest of having consistent and readable SQL files, this commit adds sqlfluff linting of .sql files to the npm lint script.


package.json
```diff
@@ -7,7 +7,8 @@
     "check-types": "tsc",
     "check-format": "prettier --check src",
     "eslint": "eslint --max-warnings 0 ./src --ext .ts",
-    "lint": "npm run check-format && npm run check-types && npm run eslint",
+    "lint-sql": "docker run -it --rm -v $PWD:/sql sqlfluff/sqlfluff:2.3.5 lint --dialect postgres --nocolor src/*/queries/ src/fixtures",
```
liam-lloyd, line 10: Uses docker so we don't have to deal with local Python environments

meisekimiu (approved):

PR #66, opened 2024-01-12
Update prod EKS cluster to Kubernetes 1.25

liam-lloyd:
Kubernetes 1.24 will pass out of standard EKS support at the end of this month, and we've successfully updated our test cluster to 1.25, so this commit updates the prod cluster as well.


cecilia-donnelly (approved):
Thanks, @liam-lloyd !

PR #71, opened 2024-02-23
Fix tsserver analysis of tests

liam-lloyd:
I was observing a problem where tsserver, the standard Typescript LPS, wasn't picking up on settings in our tsconfig files for test files. This was because our tsconfig.json excluded test files in order to omit them from our builds. tsserver cannot be configured to read an alternative tsconfig file, so this commit renames the api package's tsconfig.json to tsconfig.build.json, renames its tsconfig.lint.json to tsconfig.json, and updates the build script to use tsconfig.build.json.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1967,9 +1967,9 @@
       }
     },
     "node_modules/@mailchimp/mailchimp_transactional": {
-      "version": "1.0.50",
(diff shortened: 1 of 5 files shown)


liam-lloyd:
Note that this is not merely an issue with a Vim plugin that's driving this change; `tsserver` is what VSCode uses too.

cecilia-donnelly (approved):
This works for me. I think a @meisekimiu review would be more useful though.

PR #72, opened 2024-02-24
Speed up local tests with better Docker caching

liam-lloyd:
This commit moves the step that copies the contents of packages/api onto the API's test container after the npm install step. This allows dependencies to be cached, so in cases where code has changed but dependencies haven't the test container will build much faster.


cecilia-donnelly (approved):
Nice!

PR #73, opened 2024-03-05
Add POST /event endpoint

liam-lloyd:
To avoid to tight a coupling to Mixpanel, we want our own event tracking endpoint that records events in our database and, if desired, forwards corresponding events to Mixpanel. This commit adds that endpoint. It also refactors the authentication middleware to allow for endpoints that can be authenticated as either user or admin.

### DIFF
--- a/API.md
+++ b/API.md
@@ -402,3 +402,37 @@ account's status is `status.auth.ok` and its current `subject` is `null`
   alreadyInvited: [string]
 }
 ```
+
+## Events
+
+### POST `/event`
+
+Records events in the database's `event` table. If `body.mixpanel` is populated, sends the event described by that field
+to Mixpanel.
+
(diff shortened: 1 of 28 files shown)


meisekimiu (changes requested):
We found an issue with the `ip` field on the request body of the endpoint, which is that the client can't get the user's IP (without making a call to an external endpoint anyways) and we should really just be getting it from request headers.

Additionally, we decided that the `mixpanel` object in the request body should be a bit more generalized (something like "analytics" or "3rdparty" etc) so that we don't have to change the shape of the API in the future if we don't use Mixpanel. This also means we don't have to match Mixpanel's usage of snake case in the `distinct_id` field and it can match the rest of our API style.

meisekimiu (approved):

PR #74, opened 2024-03-07
Add POST /promo endpoint

liam-lloyd:
We would like to allow admin users to create promo codes via our admin dashboard. To that end, this commit adds an endpoint for creating promo codes, accessible with admin authentication.


API.md
```diff
@@ -402,3 +402,25 @@ account's status is `status.auth.ok` and its current `subject` is `null`
...
+
+- Headers: Authorization: Bearer \<JWT from FusionAuth> (admin user)
+- Request Body
+
+```
+{
+  code: string,
+  storageInMB: number (integer > 0),
+  expirationTimestamp: string (ISO 8601 format timestamp),
+  totalUsers: number (integer > 0)
```
cecilia-donnelly, line 418: `totalUses` (an extra "r" in there, which I only noticed by copy-pasting).

cecilia-donnelly (approved):
This looks good! I was able to create one with an expiration date in the past. I'm not sure if we should disallow that. I can't imagine why we'd want to create an already-expired promo, so it might be a nice helper for people, but it's not a blocker.

One small doc fix, please, but I'll approve.

PR #77, opened 2024-03-26
Add more detail to event documentation

cecilia-donnelly:
This may be overkill, but I wanted to give @flaviuvsp one pointer for documentation on using this new (rather complicated) endpoint.

I sourced these changes largely from [this Slack thread](https://permanent-org.slack.com/archives/C01ARGVA849/p1710245671958729). Thank you for writing the initial documentation and answering questions there, @liam-lloyd !

### DIFF
--- a/API.md
+++ b/API.md
@@ -412,23 +412,32 @@ to Mixpanel.
 - Headers: Authorization: Bearer \<JWT from FusionAuth> (can be user or admin)
 - Request Body
 
+See
+[the Events Schema](https://permanent.atlassian.net/wiki/spaces/EN/pages/2118975490/Events+Schema)
+for valid values of `action` and `entity`.
+
 ```
 {
-  entity: string (see database for valid values),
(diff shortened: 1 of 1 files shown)


API.md
```diff
@@ -412,24 +412,33 @@ to Mixpanel.
 - Headers: Authorization: Bearer \<JWT from FusionAuth> (can be user or admin)
 - Request Body
 
+See
+[database](https://github.com/PermanentOrg/back-end/blob/main/library/db/migrations/20240213011017_add_event_table.sql)
```
liam-lloyd, line 416: Probably better to link to [Confluence](https://permanent.atlassian.net/wiki/spaces/EN/pages/2118975490/Events+Schema) for the list of actions and entities, since as we add new ones that migration won't reflect them.

cecilia-donnelly: Thank you! Updated.

liam-lloyd (approved):

PR #79, opened 2024-04-05
Add admin endpoint to recalculate record thumbnails

liam-lloyd:
We have some records in our database with thumbnail urls in an outdated format. We'd like to force the regeneration of those urls. This commit adds an admin endpoint that does that for a given record ID.

### DIFF
--- a/API.md
+++ b/API.md
@@ -72,6 +72,24 @@ account's status is `status.auth.ok` and its current `subject` is `null`
 }
 ```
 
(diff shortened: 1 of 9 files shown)


liam-lloyd:
The linter appears to be enforcing two contradictory rules; I'll figure out how to fix this tomorrow.

cecilia-donnelly (approved):
Thank you @liam-lloyd !

PR #81, opened 2024-04-17
Add /events/checklist endpoint

liam-lloyd:
We are working on a checklist to guide users through the initial steps of setting up and using Permanent. This commit adds an endpoint at /events/checklist to return, for each item on the checklist, whether that item has been completed.

### DIFF
--- a/packages/api/src/event/queries/get_checklist_events.sql
+++ b/packages/api/src/event/queries/get_checklist_events.sql
@@ -1,0 +1,107 @@
+WITH account_events AS (
+  SELECT event.*
+  FROM
+    event
+  INNER JOIN
+    account
+    ON
+      account.subject::uuid = event.actor_id
+  WHERE
+    account.primaryemail = :email
+)
+
+SELECT
+  EXISTS(
+    SELECT 1
+    FROM
+      account_events
+    WHERE
+      account_events.actor_type = 'user'
+      AND account_events.entity = 'archive'
+      AND account_events.action = 'create'
+  ) AS "archiveCreated",
+  EXISTS(
+    SELECT 1
+    FROM
+      account_events
+    WHERE
+      account_events.actor_type = 'user'
+      AND account_events.entity = 'account'
+      AND account_events.action = 'submit_promo'
+  ) AS "storageRedeemed",
+  EXISTS(
+    SELECT 1
+    FROM
+      account_events
+    WHERE
+      account_events.actor_type = 'user'
+      AND account_events.entity = 'legacy_contact'
+      AND account_events.action = 'create'
+  ) AS "legacyContact",
+  EXISTS(
+    SELECT 1
+    FROM
+      account_events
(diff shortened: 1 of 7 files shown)


packages/api/src/event/queries/get_checklist_events.sql
```diff
@@ -0,0 +1,106 @@
...
+  ) AS "archiveCreated",
+  EXISTS(
+    SELECT 1
+    FROM
+      account_events
+    WHERE
+      account_events.actor_type = 'user'
+      AND account_events.entity = 'account'
+      AND account_events.action = 'submit_promo'
+      AND account_events.body ->> 'promoCode' = 'TellYourStory'
```
cecilia-donnelly, line 31: Ah this might be a tough one if there are special promo codes for events, though I like the idea of looking for the first one, rather than any gift code. I do think sometimes we don't use `TellYourStory`, though. @k8lyn6 , would it work to check off the "free storage redeemed" item if they've submitted *any* promo code? Am I right that sometimes you don't use `TellYourStory`?

packages/api/src/event/queries/get_checklist_events.sql
```diff
@@ -0,0 +1,106 @@
...
+          AND (account_events.body ->> 'public')::boolean = true
+        )
+        OR
+        (
+          account_events.entity = 'folder'
+          AND account_events.action = 'copy'
+          AND (account_events.body ->> 'public')::boolean = true
+        )
+      )
+  ) AS "publishContent";
```
cecilia-donnelly, line 106: This is nice! What happens if they create an empty folder in Public? (I don't know if that appears as an empty folder in their public archive anyway.)

packages/api/src/event/queries/get_checklist_events.sql
k8lyn6: @cecilia-donnelly that's correct -- we use many different codes depending on what avenue they signed up to Permanent through (i.e. Photo Detectives, Photo Managers, etc). It would be better if this was just any code redeemed. Certainly there might be some very rare cases where they redeem a different promo that isn't their original storage, but I think the purpose of this is to prompt them to _learn_ how to redeem a code, and if they've redeemed a different code, I think that's fine (again, it's a very edge case to begin with). Thanks for working on this!

packages/api/src/event/queries/get_checklist_events.sql
liam-lloyd: An empty folder would appear in a Public archive, so arguably this should include a check for the creation of public folders. On the other hand, maybe it should be excluding moves and copies of empty folders. @k8lyn6, should adding an empty folder to the public workspace count as publishing?

k8lyn6: @liam-lloyd hmmmm. I don't think it should, but I also think it's an edge case. Let's go with no.

PR #82, opened 2024-05-04
Add script to clean up the Archivematica dashboard

liam-lloyd:
Once we start processing files with Archivematica, we will need to periodically clean out the Archivematica dashboard, to prevent it from being unusably slow if we ever log in to debug something or update configuration. This commit adds a script that cleans out all completed jobs from the dashboard. A future commit will set up this script to run on a Kubernetes CronJob workload.

### DIFF
--- a/packages/archivematica_cleanup/src/index.ts
+++ b/packages/archivematica_cleanup/src/index.ts
@@ -1,0 +1,51 @@
+import * as Sentry from "@sentry/node";
+import "./env";
+
+const env = process.env["ENV"] ?? "";
+Sentry.init({
+  dsn: process.env["SENTRY_DSN"] ?? "",
+  tracesSampleRate: 1.0,
+  environment: env === "local" ? `local-${process.env["DEV_NAME"] ?? ""}` : env,
+});
+
+export const cleanupDashboard = async (): Promise<void> => {
+  try {
+    const transferDeleteResponse = await fetch(
+      `${process.env["ARCHIVEMATICA_BASE_URL"] ?? ""}/api/transfer/delete`,
+      {
+        method: "DELETE",
+        headers: {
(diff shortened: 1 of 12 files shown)


packages/archivematica_cleanup/src/index.ts
```diff
@@ -0,0 +1,51 @@
...
+Sentry.init({
+  dsn: process.env["SENTRY_DSN"] ?? "",
+  tracesSampleRate: 1.0,
+  environment: env === "local" ? `local-${process.env["DEV_NAME"] ?? ""}` : env,
+});
+
+export const cleanupDashboard = async (): Promise<void> => {
+  try {
+    const transferDeleteResponse = await fetch(
+      `${process.env["ARCHIVEMATICA_BASE_URL"] ?? ""}/api/transfer/delete`,
```
cecilia-donnelly, line 14: I don't see this in [their API docs](https://www.archivematica.org/en/docs/archivematica-1.15/dev-manual/api/api-reference-archivematica/#api-reference-archivematica) - I only see deleting one at a time. Do you know where that's documented? (I just searched "delete" in the page so could easily have missed it.)

cecilia-donnelly (commented):
Since this only deletes completed transfers and ingests (or at least that's how I read it), is there a danger that we could also have the dashboard fill up with incomplete or hanging transfers/ingests?

liam-lloyd:
> is there a danger that we could also have the dashboard fill up with incomplete or hanging transfers/ingests?

Yes, but my hope is that when things fail we end up looking into those failures and deleting them from the dashboard manually when they're resolved.

cecilia-donnelly (approved):
I didn't test this, but it makes sense. It's unfortunate that the endpoints we're using aren't documented - maybe a good thing to contribute upstream in the future.

PR #83, opened 2024-05-08
Two factor api GET request

crisnicandrei:
This pull request adds the implementation of the get request which retrieves the authentication methods of a signed user. The response coming back from FusionAuth has been mapped over to be in sync with the API docs of stela.


API.md
```diff
@@ -499,3 +499,72 @@ MixPanel](https://permanent.atlassian.net/wiki/spaces/EN/pages/2086862849/Mixpan
...
+JWT.
+
+- Headers: Authorization: Bearer \<JWT from FusionAuth>
+- Response: For now, this endpoint will only return the user's
+  two-factor methods. This will be an array with 0-2 methods in it.
+
+```
+[
+  {
+    methodId: [string] 4 random letters and numbers
```
liam-lloyd, line 521: The brackets around here suggest that the type of `methodId` is an array of strings, but I suspect it's just a string.

packages/api/src/idpuser/controller.test.ts
```diff
@@ -0,0 +1,117 @@
+/** @format */
```
liam-lloyd, line 1: Nothing in `stela` should have the Prettier format pragma, because the whole thing is Prettier-formatted.

packages/api/src/idpuser/controller.ts
```diff
@@ -0,0 +1,46 @@
...
+        req.body.emailFromAuthToken
+      );
+      if (clientResponse.response.user.twoFactor.methods.length) {
+        response = clientResponse.response.user.twoFactor.methods.map(
+          ({ id, method, email, mobilePhone }) => ({
+            methodId: id,
+            method,
+            value: email || mobilePhone,
+          })
+        );
```
liam-lloyd, line 34: This logic should be in a separate function in a `service.ts` file. In `stela`, controllers handle request validation and interactions with the actual Request and Response objects, any other endpoint logic goes in a service.

packages/api/src/idpuser/models.ts
```diff
@@ -0,0 +1,11 @@
+/** @format */
+
+export interface TwoFactorRequest {
+  emailFromAuthToken: string;
+}
```
liam-lloyd, line 5: This appears to only be used in tests. I would either move its definition to the test file, or just delete it and use `{emailFromAuthToken: string}` in its place in the tests.

packages/api/src/idpuser/validators.ts
```diff
@@ -0,0 +1 @@
+/** @format */
```
liam-lloyd, line 1: This file should be deleted.

crisnicandrei:
@liam-lloyd thanks! All the comments have been adressed :)

packages/api/src/idpuser/service.test.ts
```diff
@@ -0,0 +1,76 @@
...
+        },
+      },
+    });
+  });
+
+  afterEach(async () => {
+    jest.clearAllMocks();
+  });
+
+  test("should return an array with the length equal to 1 is the user has one multi factor method enabled", async () => {
```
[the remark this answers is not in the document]
crisnicandrei: @cecilia-donnelly should i just remove the file or test something else?

packages/api/src/idpuser/service.test.ts
```diff
@@ -0,0 +1,39 @@
...
+describe("/idpuser", () => {
+  const mockedFusionAuthClient = fusionAuthClient as jest.Mocked<
+    typeof fusionAuthClient
+  >;
+
+  afterEach(async () => {
+    jest.clearAllMocks();
+  });
+
+  it("should correctly map values when there are two-factor methods", async () => {
```
liam-lloyd, line 15: `stela` prefers `test` over `it`. Also there should probably be a test of what happens when `clientResponse.response.user.twoFactor.methods.length` is falsy.

packages/api/src/idpuser/controller.test.ts
```diff
@@ -1,15 +1,17 @@
...
+interface TwoFactorRequest {
+  emailFromAuthToken: string;
+}
 
 jest.mock("node-fetch", () => jest.fn());
 jest.mock("../middleware");
-jest.mock("../fusionauth");
+jest.mock("./service");
 
 describe("/idpuser", () => {
```
liam-lloyd, line 16: There should be a test in here covering the case where `getTwoFactorMethods` throws an error (in particular, I'm thinking of the case where the FusionAuth call fails unexpectedly).

liam-lloyd (commented):
Now that tests covering the service functionality are going through the controller, the controller tests shouldn't mock the service. Instead, they should mock fusionauth, so functionality in the service is still tested.

crisnicandrei:
@liam-lloyd all the tests should mock fusionauth?

liam-lloyd:
> @liam-lloyd all the tests should mock fusionauth?

Yes, because it's an external service, so not our responsibility to test.

crisnicandrei:
@liam-lloyd pushed the changes

liam-lloyd (approved):

PR #84, opened 2024-05-10
Include device data in Mixpanel events

liam-lloyd:
The experience team wants to be able to segment Mixpanel events by device properties (mobile vs. desktop, browser, OS), so this commit updates POST /event to extract these properties from the user agent and include them in the Mixpanel event.

### DIFF
--- a/packages/api/src/event/service.ts
+++ b/packages/api/src/event/service.ts
@@ -1,4 +1,5 @@
 import createError from "http-errors";
+import { UAParser } from "ua-parser-js";
 import { logger } from "@stela/logger";
 import { db } from "../database";
 import { mixpanelClient } from "../mixpanel";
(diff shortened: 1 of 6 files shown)


packages/api/src/event/service.ts
```diff
@@ -10,10 +11,14 @@ import {
...
 export const createEvent = async (data: CreateEventRequest): Promise<void> => {
   if (data.body.analytics) {
+    const { browser, os, device } = UAParser(data.userAgent);
     try {
       const analyticsData: { [key: string]: unknown; distinct_id?: string } =
         data.body.analytics.data;
       analyticsData.distinct_id = data.body.analytics.distinctId;
+      analyticsData["$browser"] = browser.name;
+      analyticsData["$os"] = os.name;
+      analyticsData["$device"] = device.type;
```
cecilia-donnelly, line 21: Does Mixpanel require these names?

cecilia-donnelly (approved):
This looks good! Thanks! I was curious about the dollar signs in the keys - is that a MixPanel requirement?

packages/api/src/event/service.ts
liam-lloyd: Yes, I think because its UI surfaces these fields separately from the event body in general

PR #86, opened 2024-06-10
Add /idpuser/enable-two-factor endpoint

liam-lloyd:
We want to give users the ability to manage their 2FA settings via the Permanent clients. To that end, we need an endpoint for enabling 2FA. This commit adds such an endpoint.

### DIFF
--- a/packages/api/src/idpuser/validators.ts
+++ b/packages/api/src/idpuser/validators.ts
@@ -1,5 +1,8 @@
 import Joi from "joi";
-import type { SendEnableCodeRequest } from "./models";
+import type {
+  SendEnableCodeRequest,
+  CreateTwoFactorMethodRequest,
+} from "./models";
 
 export function validateSendEnableCodeRequest(
   data: unknown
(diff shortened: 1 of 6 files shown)


packages/api/src/idpuser/validators.ts
```diff
@@ -0,0 +1,21 @@
...
+import type { CreateTwoFactorMethodRequest } from "./models";
+
+export function validateCreateTwoFactorMethodRequest(
+  data: unknown
+): asserts data is CreateTwoFactorMethodRequest {
+  const validation = Joi.object()
+    .keys({
+      emailFromAuthToken: Joi.string().email().required(),
+      code: Joi.string().required(),
+      method: Joi.string().valid("email", "sms", "authenticator").required(),
```
cecilia-donnelly, line 11: Authenticator is valid but we're not allowing it at the moment, so I would probably leave it out here.

packages/api/src/types/fusionauth.d.ts
```diff
@@ -45,6 +45,30 @@ declare module "@fusionauth/typescript-client" {
...
         };
       };
+      exception: Error;
+      wasSuccessful: () => boolean;
+    }>;
+    public enableTwoFactor(
+      userId: string,
+      request: {
+        applicationId?: UUID;
+        authenticatorId?: string;
```
cecilia-donnelly, line 55: Same note about this and the secret fields below.

cecilia-donnelly (approved):

PR #92, opened 2024-06-26
Parse list of IPs properly in `extractIp`

meisekimiu:
The back-end will pass a list of proxies from the X-Forwarded-For header, [if needed](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#syntax
). The `extractIp` middleware was not written to expect this format, and so would cause validation errors when passed a list of proxy IPs. Update this middleware function to handle a proxy IP list by extracting out the client IP from the list.

Resolves PER-9651: "Publish archive" does not get checked off in checklist

### DIFF
--- a/packages/api/src/middleware/extract_ip.test.ts
+++ b/packages/api/src/middleware/extract_ip.test.ts
@@ -2,11 +2,42 @@ import type { Request, Response } from "express";
 import { extractIp } from "./extract_ip";
 
 describe("extractIp", () => {
-  test("should extract the IP from a request", () => {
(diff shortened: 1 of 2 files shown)


meisekimiu:
I have a version of this branch deployed to dev right now, and not only does it fix the "Publish archive" checklist item, but I do believe this also fixes the "create archive" one as well, which is great because I wasn't even *trying* to fix that one! Feel free to test yourself, and compare to staging or a dev environment with the current main branch deployed.

liam-lloyd (approved):

PR #95, opened 2024-07-01
Update EKS clusters

liam-lloyd:
Kubernetes v1.26 is already on extended support in EKS, which costs more. To avoid that extra cost, this commit updates the clusters to v1.30, the latest Kubernetes version.

### DIFF
--- a/terraform/prod_cluster/eks-cluster.tf
+++ b/terraform/prod_cluster/eks-cluster.tf
@@ -3,7 +3,7 @@ module "eks" {
   version = "19.0.4"
 
   cluster_name    = local.cluster_name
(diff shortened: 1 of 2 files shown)


cecilia-donnelly (approved):
Thank you! You had mentioned that you thought this would be a tougher change the previous one - I take it that's not how it turned out?

liam-lloyd:
> Thank you! You had mentioned that you thought this would be a tougher change the previous one - I take it that's not how it turned out?

Yeah, I read something when I did the previous upgrade that seemed at the time to suggest that the next one might actually impact us, but that doesn't seem to be the case.

PR #96, opened 2024-07-02
Add SSL certificates for AWS RDS to deployed containers

liam-lloyd:
Presently, we don't use SSL in our database connections. This is a security best practice, so we should. This commit adds the public key certificates for the us-west-2 RDS signing authorities to the deployed API containers, so they can be specified as trusted signing authorities when we connect to the database.


liam-lloyd:
To answer a question I might have if I was reviewing this, the PostgreSQL client gets the cert file from its connection string, which is an environment variable in Terraform. I'll update that after this is merged.

cecilia-donnelly (approved):
Thanks! I wouldn't have caught the Postgres client, but thanks for explaining that as well.

PR #99, opened 2024-07-11
Add New Relic monitoring for dev API

liam-lloyd:


Dockerfile.api
```diff
@@ -34,4 +38,4 @@ RUN npm install --workspace @stela/api
 ENV PORT=80
 EXPOSE 80
 
-CMD node packages/api/dist/index.js
+CMD node -r newrelic packages/api/dist/index.js
```
valentinvsp, line 41: i'm curious about the use of the cli require flag. did you have issues importing newrelic from node_modules?

valentinvsp (approved):

Dockerfile.api
liam-lloyd: That was left over from some troubleshooting I was doing; it did not turn out to be necessary. Thanks for catching it!

PR #102, opened 2024-07-19
PER-8805 Allow remove self from archive

valentinvsp:


packages/api/src/account/queries/get_account_archive.sql
```diff
@@ -0,0 +1,15 @@
...
+  account_archive.status
+FROM
+  account_archive
+INNER JOIN
+  account
+  ON account_archive.accountid = account.accountid
+WHERE
+  account_archive.archiveid = :archiveId
+  AND account.primaryemail = :email
+  AND account_archive.status = 'status.generic.ok';
```
cecilia-donnelly, line 15: Note to self: this probably makes sense, but I wonder if we want to use this for declining an archive invitation? I should investigate how that works right now.

cecilia-donnelly: (Relevant because that would be in status "pending")

packages/api/src/account/service.test.ts
```diff
@@ -147,3 +152,110 @@ describe("getSignupDetails", () => {
...
+
+  afterEach(async () => {
+    await clearDatabase();
+  });
+
+  test("should successfully leave an archive", async () => {
+    const selectAccountArchiveRow = `SELECT * FROM account_archive WHERE
+    accountid = 3 AND archiveid = 1`;
+
+    const accounArchiveBeforeLeaveResult = await db.query(
```
cecilia-donnelly, line 180: Typo on this variable name - should be `accountArchiveBeforeLeaveResult`

packages/api/src/account/service.test.ts
```diff
@@ -147,3 +152,110 @@ describe("getSignupDetails", () => {
...
+    expect(accounArchiveBeforeLeaveResult.rows.length).toBe(1);
+
+    await accountService.leaveArchive({
+      emailFromAuthToken: "test+1@permanent.org",
+      userSubjectFromAuthToken: "b5461dc2-1eb0-450e-b710-fef7b2cafe1e",
+      ip,
+      archiveId: "1",
+    });
+
+    const accounArchiveAfterLeaveResult = await db.query(
```
cecilia-donnelly, line 192: Same typo (missing "t").

packages/api/src/account/validators.test.ts
```diff
@@ -1,6 +1,10 @@
-import { validateUpdateTagsRequest } from "./validators";
+import {
+  validateLeaveArchiveParams,
+  validateLeaveArchiveRequest,
+  validateUpdateTagsRequest,
+} from "./validators";
 
-describe("validateUpdateTagsRequest", () => {
+describe.skip("validateUpdateTagsRequest", () => {
```
cecilia-donnelly, line 7: Probably we don't want to skip these?

packages/api/src/account/validators.test.ts
```diff
@@ -125,3 +129,138 @@ describe("validateUpdateTagsRequest", () => {
...
+        archiveId: 123,
+      });
+    } catch (err) {
+      error = err;
+    } finally {
+      expect(error).not.toBeNull();
+    }
+  });
+
+  test("should error if archiveId string contains non-numeric characters", () => {
```
cecilia-donnelly, line 171: The name of this is a little confusing to me since the test value is all numeric.

cecilia-donnelly (commented):
A more general code organization comment is that we usually write tests on the controller, not on the service. I don't think that's a blocker (and thank you for the thorough tests!) but for consistency that would be somewhat better.

cecilia-donnelly:
We also merged our big mobbing change today that probably caused the conflicts on this PR - sorry about that!

API.md
```diff
@@ -322,6 +322,14 @@ Queues a thumbnail regeneration task for the record given by `recordId`
 }
 ```
 
+### DELETE `/account/archive/:archiveId`
+
+Used by accounts that want to leave an archive that is shared with them.
```
liam-lloyd, line 327: I would say "leave an archive of which they are a member", just to avoid conflating terms, because "sharing" is a separate concept in our application (it applies to records and folders, not to archives).

API.md
```diff
@@ -468,6 +476,7 @@ MixPanel](https://permanent.atlassian.net/wiki/spaces/EN/pages/2086862849/Mixpan
 
 - Headers: Authorization: Bearer \<JWT from FusionAuth>
 - Response
+
```
liam-lloyd, line 479: I'm guessing these whitespace changes came from an auto-formatter; what tool are you using?

packages/api/src/account/queries/delete_account_archive.sql
```diff
@@ -0,0 +1,6 @@
+DELETE FROM account_archive
+WHERE
+  archiveid = :archiveId
+  AND accountid IN (SELECT accountid FROM account WHERE primaryemail = :email)
```
liam-lloyd, line 4: For some added defensiveness, I'd include a check that this `account_archive` isn't an ownership relationship.

packages/api/src/account/service.ts
```diff
@@ -50,7 +60,72 @@ const getSignupDetails = async (
...
+      action: EVENT_ACTION.Update,
+      entity: EVENT_ENTITY.Account,
+      entityId: accountArchive.accountId.toString(),
+      ip,
+      version: 1,
+      body: {
+        archiveId,
+        accountArchiveId: accountArchive.accountArchiveId,
+      },
+      userSubjectFromAuthToken,
```
liam-lloyd, line 119: I don't think this event would communicate to the future reviewer of the event logs that this was a deletion of an `account_archive`. The way I'd communicate that would be to add an `account_archive` `entity` type and a `deleted` action type and make this an `account_archive.deleted` event.

packages/api/src/account/validators.ts
```diff
@@ -19,3 +19,35 @@ export function validateUpdateTagsRequest(
...
   }
 }
+
+export function validateLeaveArchiveParams(
+  data: unknown
+): asserts data is { archiveId: string } {
+  const validation = Joi.object()
+    .keys({
+      archiveId: Joi.string()
+        .regex(/^[1-9]\d+$/)
```
liam-lloyd, line 29: I would omit this check; while it's true that presently a string that doesn't look like a positive integer couldn't be an `archiveId`, we'd eventually like all our IDs to be UUIDs instead, so I don't want to add code that prohibits that.

packages/api/src/account/validators.ts
```diff
@@ -19,3 +19,35 @@ export function validateUpdateTagsRequest(
...
+export function validateLeaveArchiveRequest(data: unknown): asserts data is {
+  ip: string;
+  emailFromAuthToken: string;
+  userSubjectFromAuthToken: string;
+} {
+  const validation = Joi.object()
+    .keys({
+      ip: Joi.string().ip().required(),
+      emailFromAuthToken: Joi.string().email().required(),
+      userSubjectFromAuthToken: Joi.string().uuid(),
```
liam-lloyd, line 47: I think this should be required too

API.md
valentinvsp: i use Prettier for formatting. should I be using something else? or deactivate it on `.md` files?

packages/api/src/account/service.ts
valentinvsp: i think this is a valid point. perhaps we can decide how to move forward during today's planning session?

packages/api/src/account/validators.ts
valentinvsp: i decided to allow both numeric and uuid values, to have some sort of check in place.

packages/api/src/account/service.ts
cecilia-donnelly: We discussed and will do this - @valentinvsp will open a PR in backend to add those new values to the database.

API.md
liam-lloyd: Prettier is our formatter of choice, and I see no reason that it shouldn't be used on Markdown files. I was curious because I thought my Prettier checks should have caught this, but I realize now that my editor isn't currently running it on Markdown files; I'll change that.

liam-lloyd (approved):
Looks like the tests will pass once the DB migration from back-end is merged, so this looks good to me!

cecilia-donnelly (approved):
Thank you! I tested and this worked great.

cecilia-donnelly:
@valentinvsp our process is usually that you merge your own PR once it's been approved.

PR #104, opened 2024-07-23
Fix Sentry logging from Archivematica cleanup cron

liam-lloyd:
The Archivematica cleanup cron is supposed to send error logs to Sentry, but it currently doesn't. This seems to be because we needed to include an extra Sentry package; this commit adds it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3273,9 +3273,9 @@
       }
     },
     "node_modules/@types/connect": {
-      "version": "3.4.33",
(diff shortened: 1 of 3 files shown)


liam-lloyd:
This carries some extra urgency because we have an issue where this cron appears to be broken only in prod, and this PR should help us figure out why.

valentinvsp (approved):

cecilia-donnelly (approved):

PR #109, opened 2024-07-30
Add GET /promo endpoint

liam-lloyd:
We are working on building a tool in our admin dashboard so the Experience Team can manage promo codes. This tool should be able to display all existing promo codes, so we need an endpoint that will return them all. This commit adds such and endpoint.

### DIFF
--- a/packages/api/src/promo/queries/get_promos.sql
+++ b/packages/api/src/promo/queries/get_promos.sql
@@ -1,0 +1,12 @@
+SELECT
+  promoid AS "id",
+  code,
+  sizeinmb AS "storageInMB",
+  expiresdt AS "expirationTimestamp",
+  remaininguses AS "remainingUses",
+  status,
+  type,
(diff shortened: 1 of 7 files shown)


packages/api/src/promo/queries/get_promos.sql
```diff
@@ -0,0 +1,12 @@
+SELECT
+  promoid AS "id",
+  code,
+  sizeinmb AS "storageInMB",
+  expiresdt AS "expirationTimestamp",
+  remaininguses AS "remainingUses",
+  status,
+  type,
+  createddt AS "createdAt",
+  updateddt AS "updatedAt"
```
liam-lloyd, line 10: I wasn't certain whether `promoid`, `status`, and `type` should be returned here, as all are fairly internal (`status` and `type` are functionally unused). I decided to err on the side of exposing things since this is an admin endpoint.

PR #110, opened 2024-08-05
Per 9772 fix stela request validation

liam-lloyd:


packages/api/src/validators/shared.ts
```diff
@@ -2,9 +2,15 @@ import Joi from "joi";
...
-): asserts data is { emailFromAuthToken: string } {
+): asserts data is {
+  emailFromAuthToken: string;
+  userSubjectFromAuthToken: string;
+} {
   const validation = Joi.object()
-    .keys({ emailFromAuthToken: Joi.string().email().required() })
+    .keys({
+      emailFromAuthToken: Joi.string().email().required(),
+      userSubjectFromAuthToken: Joi.string().required(),
```
cecilia-donnelly: Joi has a "uuid" subtype of string that we should use, I think (though of course that means changing the test strings elsewhere).

cecilia-donnelly (approved):
This looks right to me! Thanks @liam-lloyd for putting this together.

PR #111, opened 2024-08-06
Send IPs to Mixpanel

liam-lloyd:
We want to track locations in Mixpanel, and these are determined based on IP. Therefore, we need to pass caller IPs to Mixpanel; this commit does that.


liam-lloyd:
Note that testing this on local doesn't seem to work, because the caller IP in this case will be a private IP from which Mixpanel can't retrieve a location. You can see it work on dev.

cecilia-donnelly (approved):
Thank you @liam-lloyd !

PR #112, opened 2024-08-06
Fix terraform formatting error

liam-lloyd:


meisekimiu (commented):
Thanks for fixing this!

PR #114, opened 2024-08-14
PER-9743: Add build stage to Dockerfile for local dev env

valentinvsp:
Included `npm run build -ws` step because the running api package references files in the dist folders of the other packages.


cecilia-donnelly (approved):
This looks fine to me.

liam-lloyd (approved):

PR #115, opened 2024-08-16
Add lambda handler for attaching thumbnails to records

liam-lloyd:
Now that Archivematica is generating thumbnails for us, we want to use those instead of generating our own. This commit adds a lambda handler that will be subscribed to S3 events fired when Archivematica generates new .jpg files inside DIPs. The handler checks whether the .jpg is a thumbnail, and if so creates a CDN URL for it and adds that URL, and the S3 location of the thumbnail, to the corresponding record in our database.


packages/record_thumbnail_attacher/src/index.test.ts
```diff
@@ -0,0 +1,381 @@
...
+    await loadFixtures();
+  });
+
+  afterEach(async () => {
+    await clearDatabase();
+    jest.clearAllMocks();
+    jest.restoreAllMocks();
+  });
+
+  test("should take no action if the file isn't a thumbnail", async () => {
```
liam-lloyd, line 44: This test doesn't test anything, it should check that the record isn't updated.

meisekimiu (approved):

PR #117, opened 2024-09-03
Add infrastructure as code for record thumbnail lambda

liam-lloyd:
Because upload volume is highly variable, we would like to use a lambda to attach thumbnails created by Archivematica to new records in our database. This commit adds the necessary terraform configuration to deploy such a lambda, and updates the deploy actions of the repo to include the lambda in deploys. It also updates the lambda handler to correctly reflect the format of the messages that will trigger the lambda.

### DIFF
--- a/.github/workflows/dev_deploy.yml
+++ b/.github/workflows/dev_deploy.yml
@@ -9,16 +9,19 @@ jobs:
     uses: ./.github/workflows/test.yml
     secrets: inherit
 
+  generate_image_tags:
+    uses: ./.github/workflows/generate_image_tags.yml
+    secrets: inherit
+
   build_api:
     needs:
       - run_tests
+      - generate_image_tags
     runs-on: ubuntu-20.04
+    env:
+      API_IMAGE_TAG: ${{ needs.generate_image_tags.outputs.API_IMAGE_TAG }}
     steps:
       - uses: actions/checkout@v3
-      - name: Generate Image Tag
-        run: echo "API_IMAGE_TAG=364159549467.dkr.ecr.$AWS_REGION.amazonaws.com/stela:api-$([[ ${GITHUB_REF##*/} = main ]] && echo main || echo feature)-$(git rev-parse --short $GITHUB_SHA)" >> $GITHUB_ENV
-        env:
-          AWS_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
       - name: Build Image
         run: docker build -t $API_IMAGE_TAG --build-arg="AWS_RDS_CERT_BUNDLE=$AWS_RDS_CERT_BUNDLE" -f Dockerfile.api .
         env:
@@ -34,13 +37,12 @@ jobs:
   build_am_cleanup:
     needs:
       - run_tests
+      - generate_image_tags
     runs-on: ubuntu-20.04
+    env:
+      AM_CLEANUP_IMAGE_TAG: ${{ needs.generate_image_tags.outputs.AM_CLEANUP_IMAGE_TAG }}
     steps:
       - uses: actions/checkout@v3
-      - name: Generate Image Tag
-        run: echo "AM_CLEANUP_IMAGE_TAG=364159549467.dkr.ecr.$AWS_REGION.amazonaws.com/stela:am_cleanup-$([[ ${GITHUB_REF##*/} = main ]] && echo main || echo feature)-$(git rev-parse --short $GITHUB_SHA)" >> $GITHUB_ENV
-        env:
-          AWS_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
       - name: Build Image
         run: docker build -t $AM_CLEANUP_IMAGE_TAG -f Dockerfile.am_cleanup .
       - name: AWS Login
@@ -54,13 +56,12 @@ jobs:
   build_record_thumbnail_lambda:
     needs:
       - run_tests
+      - generate_image_tags
     runs-on: ubuntu-20.04
+    env:
+      RECORD_THUMBNAIL_LAMBDA_IMAGE_TAG: ${{ needs.generate_image_tags.outputs.RECORD_THUMBNAIL_LAMBDA_IMAGE_TAG }}
     steps:
       - uses: actions/checkout@v3
-      - name: Generate Image Tag
-        run: echo "RECORD_THUMBNAIL_LAMBDA_IMAGE_TAG=364159549467.dkr.ecr.$AWS_REGION.amazonaws.com/stela:record_thumbnail-$([[ ${GITHUB_REF##*/} = main ]] && echo main || echo feature)-$(git rev-parse --short $GITHUB_SHA)" >> $GITHUB_ENV
-        env:
-          AWS_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
       - name: Build Image
         run: docker build -t $RECORD_THUMBNAIL_LAMBDA_IMAGE_TAG --build-arg="AWS_RDS_CERT_BUNDLE=$AWS_RDS_CERT_BUNDLE" -f Dockerfile.record_thumbnail_attacher .
         env:
@@ -78,7 +79,12 @@ jobs:
       - build_api
       - build_am_cleanup
       - build_record_thumbnail_lambda
+      - generate_image_tags
     runs-on: ubuntu-20.04
+    env:
+      API_IMAGE_TAG: ${{ needs.generate_image_tags.outputs.API_IMAGE_TAG }}
+      AM_CLEANUP_IMAGE_TAG: ${{ needs.generate_image_tags.outputs.AM_CLEANUP_IMAGE_TAG }}
+      RECORD_THUMBNAIL_LAMBDA_IMAGE_TAG: ${{ needs.generate_image_tags.outputs.RECORD_THUMBNAIL_LAMBDA_IMAGE_TAG }}
     defaults:
       run:
         working-directory: ./terraform/test_cluster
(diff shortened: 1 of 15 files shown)


meisekimiu (approved):

.github/workflows/staging_deploy.yml
```diff
@@ -48,10 +48,33 @@ jobs:
...
       - name: Publish Image to ECR
         run: docker push $AM_CLEANUP_IMAGE_TAG
+  build_record_thumbnail_lambda:
+    needs:
+      - run_tests
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+      - name: Generate Image Tag
+        run: echo "RECORD_THUMBNAIL_LAMBDA_IMAGE_TAG=364159549467.dkr.ecr.$AWS_REGION.amazonaws.com/stela:record_thumbnail-$([[ ${GITHUB_REF##*/} = main ]] && echo main || echo feature)-$(git rev-parse --short $GITHUB_SHA)" >> $GITHUB_ENV
```
slifty, line 58: There is a lot happening in one line here -- would it be possible to write this in a way that visually organizes the granular steps (even if it's all one command in the end)

Dockerfile.record_thumbnail_attacher
```diff
@@ -14,14 +14,19 @@ RUN npm run build -ws
...
 
+RUN mkdir /etc/ca-certificates
+RUN echo -e $AWS_RDS_CERT_BUNDLE > /etc/ca-certificates/rds-us-west-2-ca-bundle.pem
+
 COPY --from=builder /usr/local/apps/stela/packages/record_thumbnail_attacher/dist ./packages/record_thumbnail_attacher/dist
 COPY --from=builder /usr/local/apps/stela/packages/record_thumbnail_attacher/package.json ./packages/record_thumbnail_attacher/package.json
 COPY --from=builder /usr/local/apps/stela/packages/logger/dist ./packages/logger/dist
 COPY --from=builder /usr/local/apps/stela/packages/logger/package.json ./packages/logger/package.json
 COPY --from=builder /usr/local/apps/stela/package.json ./package.json
-COPY --from=builder /usr/local/apps/stela/package-lock.json ./package-lock.json
```
slifty, line 24: What was the reason for removing this from the build step?  Depending on the answer it might be worth including in the commit description (or, potentially, having its own commit to explain the reasoning!)

packages/record_thumbnail_attacher/src/index.test.ts
```diff
@@ -48,11 +48,17 @@ describe("handler", () => {
           messageId: "1",
           receiptHandle: "1",
           body: JSON.stringify({
-            s3: {
-              object: {
-                key: "_Liam/access_copies/e38e/8582/b417/430c/953d/5c7e/8040/1ae2/2_upload-cb45fa84-f0ea-4a9e-b1da-309e485a4f4a/object/710a1def-caf8-48f2-8eee-0848b4cfda10.jpg",
-              },
-            },
+            Message: JSON.stringify({
```
slifty, line 51: Are these test change related to the infra as code changes or is it fixing a separate issue?

It may be worth putting them in their own commit just to contextualize the reason for the change.

terraform/test_cluster/variables.tf
```diff
@@ -264,3 +274,28 @@ variable "staging_new_relic_app_name" {
   type        = string
   default     = "stela-api-staging"
 }
+
+variable "dev_cloudfront_url" {
```
slifty, line 278: I am *not* a terraform expert by any means -- do you know if we need to repeat these variable definitions across test / prod cluster, or might there be a pattern for sharing this kind of variable that is universal to all environments / clusters.

slifty (changes requested):
I raised a few bigger-picture questions that would be great to talk through (and please note that the conclusion of that discussion can very well be "nah!')

Marking this as "request changes" just because it would be good to have that discussion before hitting the merge button, even if we end up deciding not to implement changes to this PR directly.

liam-lloyd:
> I also get the sense that there is more than one commit's worth of work in this PR (but may very well be wrong wrong wrong). For instance, the changes that wrap various calls with a Record seems like it might be unrelated to the terraform work? If that's the case, let's split it into multiple commits with appropriate descriptions to help out with the git history.

The message format changes are indeed unrelated to the terraform work; I've divided them into separate commits.

.github/workflows/dev_deploy.yml
```diff
@@ -51,10 +51,33 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
       - name: Publish Image to ECR
         run: docker push $AM_CLEANUP_IMAGE_TAG
+  build_record_thumbnail_lambda:
```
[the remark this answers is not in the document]
liam-lloyd: I do think we want each lambda to have a separate build step; the O(n) pipeline steps don't worry me since the build steps can (and do) run in parallel. As long as each deployable object in this repo gets its own image, this seems like the way to do things.
In considering this, though, I have come to the realization the *reason* each deployable object in this repo gets its own image is a misunderstanding on my part. The first time I started working on a lambda in this project (about a year ago) I read that the code of a lambda had a size limit of 50 MB. I was concerned that in Typescript a monolithic project might exceed this, so I divided `stela` into multiple packages that could be deployed separately using npm workspaces. As it turns out though, that 50 MB limit only applies if you upload your code via zip file; if you use docker images they can be up to 10 GB. So while I originally though that a single application capable or running any of our lambdas or our api server or anything else in the project was off the table, I think I was mistaken. That said, the decoupling benefits of the current approach may be worthwhile on their own merits.

.github/workflows/dev_deploy.yml
```diff
@@ -51,10 +51,33 @@ jobs:
...
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      - name: Publish Image to ECR
+        run: docker push $RECORD_THUMBNAIL_LAMBDA_IMAGE_TAG
   deploy:
     needs:
       - build_api
       - build_am_cleanup
+      - build_record_thumbnail_lambda
```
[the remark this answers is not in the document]
liam-lloyd: I agree it would be better to restrict what we build/deploy to what was actually changed. If I understand correctly, the path filter only works at the workflow level though, so we'd need to create separate workflows for each buildable thing. I don't think there's anything wrong with that, but it probably fits into the refactor suggested in the next comment more than it fits into this PR.

.github/workflows/full_test_deploy.yml
```diff
@@ -48,10 +48,33 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
       - name: Publish Image to ECR
         run: docker push $AM_CLEANUP_IMAGE_TAG
+  build_record_thumbnail_lambda:
```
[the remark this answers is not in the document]
liam-lloyd: This issue occurred to me while I was working on this, and I created a [ticket](https://permanent.atlassian.net/browse/PER-9799) to at least turn the build steps into a reusable workflow. I think the environments + reusable workflows approach is the right one to take here, and we are actually using both of those features in some places already (the tests workflow is reusable, and we use environments to add protections to prod deploys). I think this might be a good project for me to undertake in the potholes sprint in the beginning of next month.

.github/workflows/dev_deploy.yml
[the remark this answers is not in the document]
slifty: Fully agree with all of that analysis!

.github/workflows/staging_deploy.yml
liam-lloyd: This is now broken into several pieces, which should make things clearer. Also, image tag generation is now a reusable workflow.

Dockerfile.record_thumbnail_attacher
liam-lloyd: I pulled this out when I was trying to figure out what I thought was a dependency versioning issue. That issue turned out to be something else, so I've put this back in.

terraform/test_cluster/variables.tf
liam-lloyd: I've looked a bit, and I haven't found a way to do this for the variables files where the variables are declared. There is a way to do this with the variable _assignments_ in Terraform Cloud, but even that has limited utility for us, because of the way our test environments are in the same workspace (so usually there aren't duplicate variables across workspaces, the test one will have `dev_cloudfront_url` and `staging_cloudfront_url`, while the prod one will just have `cloudfront_url`). The reason the workspaces are organized this way is so we can save money by running staging and dev on the same EKS cluster.

.github/workflows/generate_image_tags.yml
```diff
@@ -0,0 +1,39 @@
...
+      RECORD_THUMBNAIL_LAMBDA_IMAGE_TAG:
+        value: ${{ jobs.generate_image_tags.outputs.RECORD_THUMBNAIL_LAMBDA_IMAGE_TAG }}
+jobs:
+  generate_image_tags:
+    runs-on: ubuntu-20.04
+    outputs:
+      API_IMAGE_TAG: ${{ steps.generate_api_image_tag.outputs.API_IMAGE_TAG }}
+      AM_CLEANUP_IMAGE_TAG: ${{ steps.generate_am_cleanup_image_tag.outputs.AM_CLEANUP_IMAGE_TAG }}
+      RECORD_THUMBNAIL_LAMBDA_IMAGE_TAG: ${{ steps.generate_record_thumbnail_lambda_image_tag.outputs.RECORD_THUMBNAIL_LAMBDA_IMAGE_TAG }}
+    steps:
```
slifty, line 18: This whole section is spectacular and self documenting, very nice!

slifty (approved):
This is really great work; and thank you for all of the thoughtful iteration!

I should note that I did not *run* any of this and this is a review based on code review rather than testing, but it's all looking good.

PR #118, opened 2024-09-04
Expect Admin Subject on Promo endpoints

meisekimiu:
Previously these endpoints were configured to expect a `userSubjectFromAuthToken`, which is not assigned because these are admin endpoints. Eventually we want to get admin subjects in some endpoints, so add `adminSubjectFromAuthToken` in admin verification middleware and shared admin authentication fields.

It doesn't seem like this issue affects other endpoints as far as I can tell, but I'd appreciate another check during review.

### DIFF
--- a/packages/api/src/admin/controller.test.ts
+++ b/packages/api/src/admin/controller.test.ts
@@ -26,13 +26,14 @@ describe("recalculateFolderThumbnails", () => {
   beforeEach(async () => {
     (verifyAdminAuthentication as jest.Mock).mockImplementation(
       (req: Request, __, next: NextFunction) => {
-        (
-          req.body as {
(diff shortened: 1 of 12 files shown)


liam-lloyd (changes requested):
This will affect other admin-authenticated endpoints, because Joi fails on unexpected additional fields. The rest of the admin-authenticated endpoints should have their validators updated to draw the validations for the token fields from `fieldsFromAdminAuthentication`, like you do for `validateCreatePromoRequest` here.

meisekimiu:
I think I've updated all the admin endpoints to expect an admin subject now.

liam-lloyd (approved):

PR #119, opened 2024-09-05
Per 9737 add auth api design docs

liam-lloyd:


packages/api/docs/future/paths/authentication.yaml
```diff
@@ -0,0 +1,170 @@
...
+        application/json:
+          schema:
+            type: object
+            properties:
+              required: [email, password, applicationId]
+              email:
+                type: string
+              password:
+                type: string
+              applicationId:
```
liam-lloyd, line 19: Note that this API will eventually include endpoints for managing applications, but these are not included in this PR because we need to do some significant research and design work on those and I want to get examples of OpenAPI documentation into the repo.

package.json
```diff
@@ -23,9 +23,11 @@
     "node": ">=18.0"
   },
   "devDependencies": {
+    "@redocly/cli": "^1.22.1",
```
slifty, line 26: Oooo love it -- as an aside, in another project I am using this: https://www.npmjs.com/package/swagger-spec-validator but for a few reasons have been wanting to find an alternative.  🎉

packages/api/docs/future/api.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - url: https://api.permanent.org/v2
+components:
+  securitySchemes:
+    bearerHttpAuthentication:
+      type: http
+      scheme: bearer
+      bearerFormat: JWT
+paths:
+  /login:
+    $ref: "./paths/authentication.yaml#/login"
```
slifty, line 18: I am *loving* the way you're breaking the spec into files instead of having a single gigantic swagger specification.

Very mindful, very demure.

packages/api/docs/future/paths/authentication.yaml
```diff
@@ -0,0 +1,170 @@
...
+                  properties:
+                    authenticationToken:
+                      type: string
+                    refreshToken:
+                      description: Not yet used, but included for future use
+                      type: string
+                    account:
+                      $ref: "../models/account.yaml#/account"
+      '400':
+        $ref: "../errors.yaml#/400"
```
slifty, line 143: Referencing these reusable errors: *so good*

packages/api/package.json
```diff
@@ -21,7 +21,8 @@
     "check-format": "prettier --check src",
     "eslint": "eslint --max-warnings 0 ./src --ext .ts",
     "lint-sql": "docker run -i --rm -v $PWD:/sql sqlfluff/sqlfluff:2.3.5 lint --dialect postgres --nocolor src/*/queries/ src/fixtures",
-    "lint": "npm run check-format && npm run check-types && npm run eslint && npm run lint-sql",
+    "lint-docs": "redocly lint docs/future/api.yaml",
+    "lint": "npm run check-format && npm run check-types && npm run eslint && npm run lint-sql && npm run lint-docs",
```
slifty, line 25: This is great!

In some projects I've seen the pattern of naming sub-lint scripts as `lint:foo` (e.g. `lint:format`, `lint:types`, `lint:eslint`, `lint:sql`) and then having the scripts in alphabetical order.  This is an incredibly minor note, and I know the pattern is already established, but just to put in your ear for future in case it sounds appealing!

slifty (approved):
This PR has made me so incredibly happy, which means (1) I'm a huge nerd, but also (2) you've created something beautiful.

What an excellent patterns!

Also, I learned some new OpenAPI tricks from this work, so thank you.

Once again I did not run this locally so this is just a code review.

---

On another note, I wonder if it might be worth adding swagger-ui to this project now that you've created a spec?  This would mean stela has a publicly viewable UX for free for developers to interact with the API.  It would look [something like this](https://api.philanthropydatacommons.org).

Here's a repo that uses it: https://github.com/PhilanthropyDataCommons/service/blob/04693d1fab9c8ccc8ff275e851c0b11d2626f109/package.json#L96-L97

PR #126, opened 2024-09-21
Per 9800 add thumbnail url refresh cron

liam-lloyd:


liam-lloyd (commented):
The tests for this cron aren't currently running in CI, they should be added to the appropriate Github action

meisekimiu (approved):

PR #127, opened 2024-10-01
PER-9717 [back-end] Endpoint to retrieve environment configuration

iulianvsp:
PER-9717 [back-end] Endpoint to retrieve environment configuration

### DIFF
--- a/packages/api/src/feature_flag/controller.test.ts
+++ b/packages/api/src/feature_flag/controller.test.ts
@@ -1,0 +1,112 @@
+import request from "supertest";
+import type { NextFunction, Request } from "express";
+import { logger } from "@stela/logger";
+import { db } from "../database";
+import { app } from "../app";
+import { extractUserIsAdminFromAuthToken } from "../middleware";
+import type { FeatureFlagRequest } from "./models";
+
+jest.mock("../database");
+jest.mock("../middleware");
+jest.mock("@stela/logger");
+
+describe("GET /feature-flags", () => {
+  const agent = request(app);
+
+  const loadFixtures = async (): Promise<void> => {
+    await db.sql("fixtures.create_test_feature_flags");
+  };
+
+  const clearDatabase = async (): Promise<void> => {
+    await db.query("TRUNCATE feature_flag CASCADE");
+  };
+
+  beforeEach(async () => {
+    await clearDatabase();
+    await loadFixtures();
+    (extractUserIsAdminFromAuthToken as jest.Mock).mockImplementation(
+      (req: Request, __, next: NextFunction) => {
+        (req.body as FeatureFlagRequest).admin = false;
+        next();
+      }
+    );
+  });
+
+  afterEach(async () => {
+    jest.restoreAllMocks();
+    await clearDatabase();
+  });
+
+  test("should log the error if the database call fails", async () => {
+    const testError = new Error("SQL error");
+    jest.spyOn(db, "sql").mockRejectedValueOnce(testError);
+
+    await agent.get("/api/v2/feature-flags").expect(500);
+    expect(logger.error).toHaveBeenCalled();
+  });
+
+  test("should log the error if the database call fails when calling user is admin", async () => {
+    (extractUserIsAdminFromAuthToken as jest.Mock).mockImplementation(
+      (req: Request, __, next: NextFunction) => {
+        (req.body as FeatureFlagRequest).admin = true;
+        next();
+      }
+    );
+    const testError = new Error("SQL error");
+    jest.spyOn(db, "sql").mockRejectedValueOnce(testError);
+
+    await agent.get("/api/v2/feature-flags").expect(500);
+    expect(logger.error).toHaveBeenCalled();
+  });
+
+  test("should log the error if the request is invalid", async () => {
+    (extractUserIsAdminFromAuthToken as jest.Mock).mockImplementation(
+      (_req: Request, __, next: NextFunction) => {
+        next();
+      }
+    );
+
+    await agent.get("/api/v2/feature-flags").expect(400);
+    expect(logger.error).toHaveBeenCalled();
+  });
+
+  test("should return list of globally enabled feature flags if user is not logged in", async () => {
+    const expected = { items: [{ name: "feature_1" }] };
+
+    const response = await agent.get("/api/v2/feature-flags").expect(200);
+    expect(response.body).toEqual(expected);
+  });
+
+  test("should return list of all feature flags if user is admin", async () => {
+    (extractUserIsAdminFromAuthToken as jest.Mock).mockImplementation(
+      (req: Request, __, next: NextFunction) => {
+        (req.body as FeatureFlagRequest).admin = true;
+        next();
+      }
+    );
+
+    const expected = {
+      items: [
+        {
+          id: "1bdf2da6-026b-4e8e-9b57-a86b1817be5d",
+          name: "feature_1",
+          description: "description_1",
+          globallyEnabled: true,
+          createdAt: "2023-06-21T00:00:00.000Z",
+          updatedAt: "2023-06-21T00:00:00.000Z",
+        },
+        {
+          id: "96669879-dfbc-40c9-ba3d-00e688a158ba",
+          name: "feature_2",
+          description: "description_2",
+          globallyEnabled: false,
+          createdAt: "2023-06-21T00:00:00.000Z",
+          updatedAt: "2023-06-21T00:00:00.000Z",
+        },
+      ],
+    };
+
+    const response = await agent.get("/api/v2/feature-flags").expect(200);
+    expect(response.body).toEqual(expected);
+  });
+});
--- a/packages/api/src/feature_flag/controller.ts
+++ b/packages/api/src/feature_flag/controller.ts
@@ -1,0 +1,29 @@
+import { Router } from "express";
+import type { Request, Response, NextFunction } from "express";
+
+import { logger } from "@stela/logger";
+import { featureService } from "./service";
+import { extractUserIsAdminFromAuthToken } from "../middleware";
+import { validateIsAdminFromAuthentication } from "../validators/shared";
+import { isValidationError } from "../validators/validator_util";
+
+export const featureController = Router();
+
+featureController.get(
+  "/",
+  extractUserIsAdminFromAuthToken,
+  async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+    try {
+      validateIsAdminFromAuthentication(req.body);
+      const featureFlags = await featureService.getFeatureFlags(req.body.admin);
+      res.json({ items: featureFlags });
+    } catch (err) {
+      if (isValidationError(err)) {
+        res.status(400).json({ error: err.message });
+        return;
+      }
+      logger.error(err);
+      next(err);
+    }
+  }
+);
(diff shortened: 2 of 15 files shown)


packages/api/src/account/queries/get_feature_flags.sql
```diff
@@ -0,0 +1,6 @@
+SELECT
```
liam-lloyd, line 1: This query should live under /feature_flag/queries rather than here, since it's dealing with feature flags at the global level rather than at the account level

packages/api/src/feature/models.ts
```diff
@@ -0,0 +1,11 @@
...
+  id: number;
+  name: string;
+  globallyEnabled: boolean;
+}
+
+export interface FeatureFlagRow {
+  id: number;
+  name: string;
+  globallyEnabled: boolean;
+}
```
liam-lloyd, line 11: This looks identical to the interface above; what is it for?

packages/api/src/feature/service.ts
```diff
@@ -0,0 +1,32 @@
...
+    });
+
+  const featureFlags = result.rows.map<FeatureFlag>(
+    (row: FeatureFlagRow): FeatureFlag => ({
+      ...row,
+      id: row.id,
+      name: row.name,
+      globallyEnabled: row.globallyEnabled,
+    })
+  );
```
liam-lloyd, line 25: We use code like this in other endpoints to transform some fields where the desired output is different from what's in the database, but I don't think there's a need for that here.

packages/api/src/feature/controller.ts
```diff
@@ -0,0 +1,27 @@
...
+  verifyUserAuthentication,
+  extractIp,
+  async (_req: Request, res: Response, next: NextFunction): Promise<void> => {
+    try {
+      const featureFlags = await featureService.getFeatureFlags();
+      res.json(featureFlags);
+    } catch (err) {
+      if (isValidationError(err)) {
+        res.status(400).json({ error: err });
+        return;
```
liam-lloyd, line 22: Since this endpoint doesn't make use of any inputs, there is nothing to validate, so we needn't check for validation errors

liam-lloyd (commented):
You're generally on the right path here as far as how we write endpoints and what this one should do. I've added a couple comments in places where I think a better course exists.

You asked also whether there are any additional fields this endpoint should return; I've now added an [OpenAPI specification](https://github.com/PermanentOrg/stela/blob/main/packages/api/docs/present/paths/feature_flags.yaml) describing the desired endpoint in detail (though that specification need not be taken as dogma; if you see a better way by all means present your case for it). If you prefer not to read the YAML directly, you can create a much nicer HTML version of the specification by running `redocly build-docs packages/api/docs/present/api.yaml` (we have just recently adopted this API specification format and aren't yet hosting a docs site).

Some other documentation that might give you useful further context:
[The longer term plan for feature flags](https://permanent.atlassian.net/wiki/spaces/EN/pages/2221473793/Feature+Flags)
[Our best practices for API design](https://permanent.atlassian.net/wiki/spaces/EN/pages/2203451393/API+Design+Principles)

iulianvsp:
> You're generally on the right path here as far as how we write endpoints and what this one should do. I've added a couple comments in places where I think a better course exists.
> 
> You asked also whether there are any additional fields this endpoint should return; I've now added an [OpenAPI specification](https://github.com/PermanentOrg/stela/blob/main/packages/api/docs/present/paths/feature_flags.yaml) describing the desired endpoint in detail (though that specification need not be taken as dogma; if you see a better way by all means present your case for it). If you prefer not to read the YAML directly, you can create a much nicer HTML version of the specification by running `redocly build-docs packages/api/docs/present/api.yaml` (we have just recently adopted this API specification format and aren't yet hosting a docs site).
> 
> Some other documentation that might give you useful further context: [The longer term plan for feature flags](https://permanent.atlassian.net/wiki/spaces/EN/pages/2221473793/Feature+Flags) [Our best practices for API design](https://permanent.atlassian.net/wiki/spaces/EN/pages/2203451393/API+Design+Principles)

I updated the PR with the suggested changes, please have a look. I also saw in the best practices API design page that for lists of object we want to have pagination. Does this apply to the feature_flags as well?

Also should I create the other CRUD methods to this entity? Thanks

liam-lloyd:
> I also saw in the best practices API design page that for lists of object we want to have pagination. Does this apply to the feature_flags as well?

I think feature flags are most naturally an exception to this rule. From the user-facing client's perspective, it would be a hassle to have to make several calls through multiple pages of feature flags in order to determine which features should be active. From the admin panel's perspective, pagination might be more reasonable, but I think realistically we should never have so many feature flags at once that they would be unmanageable without pagination (unless we neglect to delete old ones, but then the lack of pagination may become a useful inducement to not neglect that).

> Also should I create the other CRUD methods to this entity?

I think that would be a great thing for you to do once this endpoint is done; the create and update methods are higher priority than delete. The tickets for those endpoints are here:
https://permanent.atlassian.net/browse/PER-9718
https://permanent.atlassian.net/browse/PER-9866

packages/api/src/feature_flag/controller.ts
```diff
@@ -0,0 +1,21 @@
...
+import type { Request, Response, NextFunction } from "express";
+
+import { featureService } from "./service";
+import { extractIp, verifyUserAuthentication } from "../middleware";
+
+export const featureController = Router();
+
+featureController.get(
+  "/features",
+  verifyUserAuthentication,
```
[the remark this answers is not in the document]
iulianvsp: I created extractUserIsAdminFromAuthToken, but please have a look at my comments there. I'm not sure if i did that correctly

packages/api/src/middleware/authentication.ts
```diff
@@ -228,9 +228,33 @@ const extractUserEmailFromAuthToken = async (
...
+      authenticationToken,
+      emailKey,
+      process.env["FUSIONAUTH_ADMIN_APPLICATION_ID"] ?? ""
+    );
+    if (email !== "") {
+      req.body.admin = true;
+    }
+    next();
+  } catch (innerErr) {
+  }
```
liam-lloyd, line 250: This empty catch doesn't seem like the right path here; in almost any case when an exception occurs we at least want to know about it. If this is what fixed the timeout error, the error is presumably still happening, it just gets swallowed by this catch now.

Since what you're doing here is just about the same as what `extractUserEmailFromAuthToken` does, I suspect what's going on is that your FusionAuth env vars aren't set to the correct values; indeed it looks like they couldn't be, because you don't have access to FusionAuth yet. I'll set up your account there now.

liam-lloyd: You should have an email inviting you to FusionAuth; once you get into their admin console you should be able to find the values you need for the environment variables.

packages/api/src/idpuser/models.ts
```diff
@@ -36,3 +36,7 @@ export enum TwoFactorMethod {
   Email = "email",
   Sms = "sms",
 }
+
+export interface FeatureFlagRequest {
+  admin: boolean;
+}
```
liam-lloyd, line 42: Shouldn't this be in the `feature_flag` models?

packages/api/src/feature_flag/controller.test.ts
```diff
@@ -0,0 +1,79 @@
+import request from "supertest";
+import type { NextFunction, Request } from "express";
+import { db } from "../database";
+import { app } from "../app";
+import { extractUserIsAdminFromAuthToken } from "../middleware";
+import type { FeatureFlagRequest } from "../idpuser/models";
+
+jest.mock("../database");
+jest.mock("../middleware");
+jest.mock("../mixpanel");
```
liam-lloyd, line 10: I don't think these tests should require mocking mixpanel

packages/api/src/feature_flag/controller.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+describe("GET /feature-flags", () => {
```
liam-lloyd, line 12: These tests look great, but it's a good idea to also test the error cases to make sure that we send the right status code and log the error correctly if, say, the database call fails.

packages/api/src/feature_flag/controller.ts
```diff
@@ -0,0 +1,22 @@
...
+import { validateIsAdminFromAuthentication } from "../validators/shared";
+
+export const featureController = Router();
+
+featureController.get(
+  "/",
+  extractUserIsAdminFromAuthToken,
+  async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+    try {
+      validateIsAdminFromAuthentication(req.body);
```
liam-lloyd, line 15: Now that we have a validator in this controller, we should add back the check for validation errors (and write a test that we return the correct status code in case of a validation error).

packages/api/src/middleware/authentication.ts
```diff
@@ -228,9 +228,25 @@ const extractUserEmailFromAuthToken = async (
   next();
 };
 
+const extractUserIsAdminFromAuthToken = async (
```
liam-lloyd, line 231: You asked about unit tests and I think my answer was incomplete. To answer more thoroughly, we favor functional tests for our controllers (and also, so far, for lambda handlers and cron scripts), because we want to ensure we test the SQL against a real database, and because separating the service from the controller in tests doesn't seem to add much value. We do generally unit test code that is shared between many controllers, such as middleware or shared validators.

liam-lloyd: It looks like this function still needs unit tests

packages/api/src/feature_flag/controller.test.ts
liam-lloyd: I'd add two more error cases: One where the database call fails when the caller is an admin, and one where the request is invalid.

packages/api/src/feature_flag/controller.test.ts
```diff
@@ -0,0 +1,112 @@
...
+  });
+
+  test("should log the error if the request is invalid", async () => {
+    (extractUserIsAdminFromAuthToken as jest.Mock).mockImplementation(
+      (_req: Request, __, next: NextFunction) => {
+        next();
+      }
+    );
+
+    await agent.get("/api/v2/feature-flags").expect(500);
```
liam-lloyd, line 69: For an invalid request we should expect a 400 error

liam-lloyd (commented):
Just one more comment, then I think this will be ready to merge!

packages/api/src/feature_flag/controller.test.ts
iulianvsp: This scenario would not be possible because extractUserIsAdminFromAuthToken would always return a boolean request.body.admin value. Maybe it would be best to just remove this test?

iulianvsp: Done, thanks :)

packages/api/src/middleware/authentication.ts
```diff
@@ -228,9 +228,25 @@ const extractUserEmailFromAuthToken = async (
...
+  req: Request<unknown, unknown, { admin?: boolean }>,
+  _: Response,
+  next: NextFunction
+): Promise<void> => {
+  const authenticationToken = getAuthTokenFromRequest(req);
+  const email = await getOptionalValueFromAuthToken(
```
liam-lloyd, line 237: I found one more thing while testing locally - this function will throw an exception (because of the way the Fusionauth client works) if it receives an `authenticationToken` that's an empty string. Since we don't want that case to fail, we should check for an empty `authenticationToken` and then only call this if it's not empty.

liam-lloyd (approved):

PR #134, opened 2024-10-09
Add controller tests for legacy contact endpoints

liam-lloyd:
Currently, the legacy contact tests don't cover the controller. This commit updates the service tests to run through the controller, and adds a few extra tests to cover invalid request errors.

PER-9515


meisekimiu (approved):
We should get codecov bot working in this repo again so we can see the effects of these test changes really easily in the pull request itself.

PR #137, opened 2024-10-22
PER-9718 [back-end] Endpoint for admins to create environment configu…

iulianvsp:
…ration

New endpoint to create a feature-flag. There are validations on the name and description + the name of the flag has to be unique.

### DIFF
--- a/packages/api/src/feature_flag/queries/create_feature_flag.sql
+++ b/packages/api/src/feature_flag/queries/create_feature_flag.sql
@@ -1,0 +1,15 @@
+INSERT INTO feature_flag (
+  name,
+  description,
+  globally_enabled
+) VALUES (
+  :name,
+  :description,
+  :globally_enabled
+) RETURNING
+id,
+name,
+description,
+globally_enabled AS "globallyEnabled",
+created_at AS "createdAt",
+updated_at AS "updatedAt";
--- a/packages/api/src/feature_flag/service/create.ts
+++ b/packages/api/src/feature_flag/service/create.ts
@@ -1,0 +1,41 @@
+import { logger } from "@stela/logger";
+import createError from "http-errors";
+import { TinyPgError } from "tinypg";
+import type { CreateFeatureFlagRequest, FeatureFlagRow } from "../models";
+import { db } from "../../database";
+
+const duplicateFeatureFlagError =
+  'duplicate key value violates unique constraint "feature_flag_name_unique"';
+
+export const createFeatureFlag = async (
+  featureFlagData: CreateFeatureFlagRequest
+): Promise<FeatureFlagRow> => {
+  const result = await db
+    .sql<FeatureFlagRow>("feature_flag.queries.create_feature_flag", {
+      name: featureFlagData.name,
+      description: featureFlagData.description,
+      globally_enabled: false,
+    })
+    .catch((err) => {
+      logger.error(err);
+      if (
+        err instanceof TinyPgError &&
+        err.message === duplicateFeatureFlagError
+      ) {
+        throw new createError.BadRequest("Feature flag already exists");
+      }
+      throw new createError.InternalServerError(
+        "Failed to create feature flag"
+      );
+    });
+
+  if (result.rows[0] === undefined) {
+    throw new createError.InternalServerError("Failed to create feature flag");
+  }
+
+  return result.rows[0];
+};
+
+export const createFeatureService = {
+  createFeatureFlag,
+};
--- a/packages/api/src/feature_flag/validators.ts
+++ b/packages/api/src/feature_flag/validators.ts
@@ -1,0 +1,19 @@
+import Joi from "joi";
+import type { CreateFeatureFlagRequest } from "./models";
+import { fieldsFromAdminAuthentication } from "../validators";
+
+export function validateCreateFeatureFlagRequest(
+  data: unknown
+): asserts data is CreateFeatureFlagRequest {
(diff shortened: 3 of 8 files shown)


packages/api/src/feature_flag/service/create.ts
```diff
@@ -0,0 +1,50 @@
...
+          "Failed to retrieve feature flags"
+        );
+      });
+
+    return result.rows.length > 0;
+  };
+
+  if (await featureFlagsWithName(featureFlagData.name)) {
+    throw new createError.BadRequest("Feature flag already exists");
+  }
```
liam-lloyd, line 26: Rather than making an extra call to the database to see whether a feature flag of the same name already exists, I'd recommend adding a unique constraint to `feature_flag.name` in the database (docs on how we do migrations: https://github.com/PermanentOrg/back-end?tab=readme-ov-file#database-migration). Then you can check for an error related to that constraint (similar to what we do here: https://github.com/PermanentOrg/back-end?tab=readme-ov-file#database-migration) on the insert operation.

packages/api/src/feature_flag/queries/create_feature_flag.sql
```diff
@@ -0,0 +1,19 @@
+INSERT INTO feature_flag (
+  name,
+  description,
+  globally_enabled,
+  created_at,
+  updated_at
```
liam-lloyd, line 6: No need to include these fields in the INSERT statement; the database's default values will be the ones we want

packages/api/src/feature_flag/service/create.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  const result = await db
+    .sql<FeatureFlagRow>("feature_flag.queries.create_feature_flag", {
+      name: featureFlagData.name,
+      description: featureFlagData.description,
+      globally_enabled: true,
```
liam-lloyd, line 32: `globally_enabled` should always be `false` when a feature flag is created, because at the moment of creation the feature flag should be turned off

packages/api/src/feature_flag/validators.ts
```diff
@@ -0,0 +1,19 @@
...
+import { fieldsFromAdminAuthentication } from "../validators";
+
+export function validateCreateFeatureFlagRequest(
+  data: unknown
+): asserts data is CreateFeatureFlagRequest {
+  const validation = Joi.object()
+    .keys({
+      ...fieldsFromAdminAuthentication,
+      name: Joi.string().required().max(255),
+      description: Joi.string().max(10000),
```
liam-lloyd, line 12: No need for the maximum lengths here; we're storing these at PostgreSQL TEXT, which does not have a character limit (technically it can't hold strings larger than 1 GB, but I'm alright with people getting a 500 error if they try to submit a > 1 GB string here)

packages/api/src/feature_flag/service/create.test.ts
```diff
@@ -0,0 +1,189 @@
+import request from "supertest";
```
liam-lloyd, line 1: Since we're testing at the controller level, this file should be at /src/feature_flag/controller.test.ts

packages/api/src/feature_flag/service/create.ts
iulianvsp: I see, ok, done. I saw it as true in the API docs and thought we want that as the default

packages/api/src/feature_flag/queries/create_feature_flag.sql
liam-lloyd: Ideally we would, but `promo` is an older table that was created without default values for anything other than the ID, so we'd have to update the database first.

packages/api/src/feature_flag/service/create.ts
liam-lloyd: Ah, yes, it appears to be the default behavior of redocly that the example value of any boolean field is `true`, which is misleading in this case. It's probably worth adding a description to the `globallyEnabled` fields in the docs noting that it always starts out as `false.

packages/api/src/feature_flag/service/create.ts
```diff
@@ -0,0 +1,41 @@
+import { logger } from "@stela/logger";
+import createError from "http-errors";
+import { TinyPgError } from "tinypg";
+import type { CreateFeatureFlagRequest, FeatureFlagRow } from "../models";
+import { db } from "../../database";
+
+const duplicateFeatureFlagError =
+  'duplicate key value violates unique constraint "feature_flag_unique"';
```
liam-lloyd, line 8: It looks like the constraint is called `feature_flag_name_unique`

packages/api/src/feature_flag/queries/create_feature_flag.sql
iulianvsp: Ah, I understand, ok.

packages/api/src/feature_flag/service/create.ts
iulianvsp: I changed it in the migration, but forgot to change it here. Nice catch, thanks :)

packages/api/src/feature_flag/queries/get_feature_flag.sql
```diff
@@ -0,0 +1,11 @@
+SELECT
```
liam-lloyd, line 1: This isn't used anymore, is it?

liam-lloyd (approved):
Approved, but do delete the unused SQL file before merging.

PR #139, opened 2024-10-25
PER-9866 [back-end] Endpoint for admins to update environment configu…

iulianvsp:
…ration

New endpoint to update a feature-flag. There are validations on the globallyEnabled and description fields

### DIFF
--- a/packages/api/src/feature_flag/controller.test.ts
+++ b/packages/api/src/feature_flag/controller.test.ts
@@ -8,7 +8,11 @@ import {
   extractUserIsAdminFromAuthToken,
   verifyAdminAuthentication,
 } from "../middleware";
-import type { CreateFeatureFlagRequest, FeatureFlagRequest } from "./models";
+import type {
+  CreateFeatureFlagRequest,
+  UpdateFeatureFlagRequest,
+  FeatureFlagRequest,
+} from "./models";
 
 jest.mock("../database");
 jest.mock("../middleware");
@@ -21,6 +25,8 @@ const loadFixtures = async (): Promise<void> => {
 const clearDatabase = async (): Promise<void> => {
   await db.query("TRUNCATE feature_flag CASCADE");
 };
+
+const notExistingFeatureFlagId = "1bdf2da6-026b-4e8e-9b57-a86b1817be4d";
 
 describe("GET /feature-flags", () => {
   const agent = request(app);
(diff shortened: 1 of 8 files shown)


packages/api/src/feature_flag/service/update.ts
```diff
@@ -0,0 +1,32 @@
...
+    })
+    .catch((err) => {
+      logger.error(err);
+      throw new createError.InternalServerError(
+        "Failed to update feature flag"
+      );
+    });
+
+  if (result.rows[0] === undefined) {
+    throw new createError.InternalServerError("Feature flag not found");
```
liam-lloyd, line 24: I think this should be a 404 error, shouldn't it?

packages/api/src/feature_flag/validators.ts
```diff
@@ -17,3 +20,32 @@ export function validateCreateFeatureFlagRequest(
...
   }
 }
+
+export function validateUpdateFeatureFlagRequest(
+  data: unknown
+): asserts data is UpdateFeatureFlagRequest {
+  const validation = Joi.object()
+    .keys({
+      ...fieldsFromAdminAuthentication,
+      description: Joi.string(),
```
liam-lloyd, line 30: `description` should be required here too, per the specification

packages/api/src/feature_flag/service/update.ts
iulianvsp: Yes, changed it, thanks

packages/api/src/feature_flag/validators.ts
iulianvsp: Ok, it seemed a bit odd not to be required on creation but required on update. I thought it's a documentation mistake.

packages/api/src/feature_flag/controller.test.ts
```diff
@@ -283,3 +287,182 @@ describe("POST /feature-flag", () => {
     expect(logger.error).toHaveBeenCalled();
   });
 });
+
+describe("PUT /feature-flag/1bdf2da6-026b-4e8e-9b57-a86b1817be5d", () => {
```
liam-lloyd, line 291: Nit: the URL param should be given as `:featureFlagId` here, rather than using a specific value

packages/api/src/feature_flag/controller.test.ts
```diff
@@ -283,3 +287,182 @@ describe("POST /feature-flag", () => {
...
+  });
+
+  test("should respond with 404 status code if feature flag does not exist", async () => {
+    await agent
+      .put("/api/v2/feature-flags/1bdf2da6-026b-4e8e-9b57-not-exist-12")
+      .send({
+        description: 1,
+        globallyEnabled: true,
+      })
+      .expect(400);
```
liam-lloyd, line 466: Should the error code here be 404?

liam-lloyd: (I think the test is passing because it's not actually reaching the 404 error, it's getting a 400 because the ID in the URL params is not a UUID)

liam-lloyd (approved):

liam-lloyd (commented):
(I clicked the wrong button, meant to comment rather than approve)

packages/api/src/feature_flag/validators.ts
liam-lloyd: Oh, you're actually right, description can't be required because this endpoint needs to be able to set `description` to NULL by not sending it. Could you correct the error in the documentation so it matches the endpoint?

liam-lloyd (approved):

PR #140, opened 2024-10-25
PER-9883 [back-end] Delete feature flag endpoint

iulianvsp:
New endpoint to delete a feature-flag

### DIFF
--- a/packages/api/src/feature_flag/controller.test.ts
+++ b/packages/api/src/feature_flag/controller.test.ts
@@ -11,6 +11,7 @@ import {
 import type {
   CreateFeatureFlagRequest,
   UpdateFeatureFlagRequest,
+  DeleteFeatureFlagRequest,
   FeatureFlagRequest,
 } from "./models";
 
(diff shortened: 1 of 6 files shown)


packages/api/src/feature_flag/queries/delete_feature_flag.sql
```diff
@@ -0,0 +1,6 @@
+DELETE FROM
+feature_flag
+WHERE
+  id = :featureFlagId
+RETURNING
+id AS "featureFlagId";
```
liam-lloyd, line 6: The linter is complaining about the whitespace at the end of this file.

packages/api/src/feature_flag/validators.ts
```diff
@@ -17,3 +17,16 @@ export function validateCreateFeatureFlagRequest(
     throw validation.error;
   }
 }
+
+export function validateDeleteFeatureFlagParams(
```
liam-lloyd, line 21: This is essentially the same as the param validator for update; perhaps they should be one function?

liam-lloyd (commented):
Could you add documentation for this endpoint to the OpenAPI specification?

packages/api/src/feature_flag/controller.test.ts
```diff
@@ -283,3 +287,82 @@ describe("POST /feature-flag", () => {
     expect(logger.error).toHaveBeenCalled();
   });
 });
+
+describe("DELETE /feature-flag/1bdf2da6-026b-4e8e-9b57-a86b1817be5d", () => {
```
liam-lloyd, line 291: I'd recommend adding a test for the 404 case

liam-lloyd (approved):

PR #141, opened 2024-10-31
PER-9719 [retool] Create "environment configuration" admin panel tool…

iulianvsp:
… in retool

Allow null values to the description field for both create and update requests.


liam-lloyd (commented):
Could you also add tests to verify that requests with null descriptions will pass these validators?

liam-lloyd (approved):

PR #142, opened 2024-11-01
Add documenation section to the README

liam-lloyd:
To make our API documentation easier to find an use, this commit adds a section to the readme describing how to do these things.


meisekimiu (approved):
At some point we should look into hosting this documentation on something like Github Pages automatically.

PR #145, opened 2024-11-18
PER-9607 [back-end] Allow clearing a location

iulianvsp:


packages/api/src/record/service.ts
```diff
@@ -25,3 +29,50 @@ export const getRecordById = async (requestQuery: {
   }));
   return records;
 };
+
+export function buildPatchQuery(
```
liam-lloyd, line 33: This needn't be exported

packages/api/src/record/service.ts
```diff
@@ -25,3 +29,50 @@ export const getRecordById = async (requestQuery: {
...
+  recordId: bigint,
+  patchRecordRequest: UpdateRecordDTO
+): string {
+  const updates: string[] = [];
+
+  // eslint-disable-next-line no-restricted-syntax
```
liam-lloyd, line 39: You can appease the linter by doing this with
```
const updates = Object.entries(patchRecordRequest)
    .filter(([_, value]) => value !== undefined)
    .map(([key, _]) => `${key} = :${key}`);
```

liam-lloyd (commented):
Could you add OpenAPI documentation for this endpoint?

packages/api/src/record/controller.test.ts
```diff
@@ -319,3 +324,96 @@ describe("record", () => {
     expect(logger.error).toHaveBeenCalledWith(testError);
   });
 });
+
+describe("patch record", () => {
```
liam-lloyd, line 328: We should have a test case covering the scenario where the update query fails at the database.

packages/api/docs/present/models/records.yaml
```diff
@@ -0,0 +1,91 @@
+record:
```
liam-lloyd, line 1: I don't think this entirely matches the model defined here https://github.com/PermanentOrg/stela/blob/c3beef495a190855ed3e4ea7162a09796b5db333/packages/api/src/record/models.ts#L1

packages/api/src/record/models.ts
```diff
@@ -97,6 +97,18 @@ export interface ShareArchive {
   name: string;
 }
 
+export interface UpdateRecordDTO {
```
liam-lloyd, line 100: I think I'd call this something like `RecordColumnsForUpdate`. I don't think DTO is universally understood, so best to avoid it in a name, and I'm also not sure this really is a Data Transfer Object, because it's not actually being sent over the network.

packages/api/src/record/service.ts
```diff
@@ -25,3 +31,40 @@ export const getRecordById = async (requestQuery: {
...
+  `;
+
+  return query.trim();
+}
+
+export const patchRecord = async (
+  recordId: string,
+  recordData: PatchRecordRequest
+): Promise<string> => {
+  const dto = requestFieldsToDatabaseFields(recordData, recordId);
```
liam-lloyd, line 58: This should renamed now that the object assigned to this variable is not called a DTO.

PR #150, opened 2024-11-28
PER-9255 Clear date on backend

iulianvsp:


packages/api/src/folder/helper.ts
```diff
@@ -0,0 +1,14 @@
+import type { PatchFolderRequest, FolderColumnsForUpdate } from "./models";
+
+export function requestFieldsToDatabaseFields(
+  request: PatchFolderRequest,
+  folderId: string
+): FolderColumnsForUpdate {
+  return {
+    archivenbr: request.archiveNumber,
```
liam-lloyd, line 8: `archivenbr` should not be updateable.

packages/api/src/folder/service.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export const patchFolder = async (
+  folderId: string,
+  folderData: PatchFolderRequest
+): Promise<string> => {
+  const columnsForUpdate = requestFieldsToDatabaseFields(folderData, folderId);
+  const query = buildPatchQuery(columnsForUpdate);
+
+  const result = await db
+    .query<FolderRow>(query, columnsForUpdate)
```
liam-lloyd, line 59: What we're getting back here isn't actually a FolderRow, is it?

packages/api/src/folder/models.ts
```diff
@@ -0,0 +1,107 @@
...
+  status: FolderStatus;
+  type: FolderType;
+  publicAt: string;
+  createdAt: string;
+  updatedAt?: string;
+  thumbnail256?: string;
+  thumbnail256CloudPath?: string;
+}
+
+export interface FolderRow {
```
liam-lloyd, line 34: What is the difference between this and `Folder`?

packages/api/src/folder/validators.ts
```diff
@@ -0,0 +1,46 @@
+import Joi from "joi";
+import type { PatchFolderRequest } from "./models";
+import { fieldsFromUserAuthentication } from "../validators";
+
+export function validateGetFolderQuery(
```
liam-lloyd, line 5: Is this used?

packages/api/src/folder/helper.ts
```diff
@@ -0,0 +1,14 @@
...
+    archiveid: request.archiveId,
```
liam-lloyd, line 9: Nor should `archiveId`

packages/api/src/folder/controller.test.ts
```diff
@@ -0,0 +1,134 @@
...
+    await clearDatabase();
+    jest.restoreAllMocks();
+    jest.clearAllMocks();
+  });
+
+  test("expect an empty query to cause a 400 error", async () => {
+    await agent.patch("/api/v2/folder/1").send({}).expect(400);
+  });
+
+  test("expect non existent record to cause a 404 error", async () => {
```
liam-lloyd, line 61: Should be "non existent folder"

packages/api/src/folder/controller.test.ts
```diff
@@ -0,0 +1,134 @@
...
+};
+
+const clearDatabase = async (): Promise<void> => {
+  await db.query(
+    `TRUNCATE 
+       folder CASCADE`
+  );
+};
+
+describe("patch folder", () => {
```
liam-lloyd, line 26: For complete coverage, we should have a test case for if the SELECT after the UPDATE fails, and another for if it returns zero rows.

packages/api/src/folder/helper.ts
iulianvsp: Should I remove archiveId as well?

packages/api/src/folder/models.ts
iulianvsp: There is no difference, I saw that we use the same approach for the ArchiveRecord and ArchiveRecordRow and used the same approach. I thought we use one as a datamodel and one for the result of the query.

packages/api/src/folder/service.ts
iulianvsp: It is a folder row object but with just the folderId property. Do you think having a new interface with with just this property is needed?

packages/api/docs/present/paths/folders.yaml
```diff
@@ -0,0 +1,37 @@
...
+      - bearerHttpAuthentication: []
+    operationId: patch-folder
+    requestBody:
+      content:
+        application/json:
+          schema:
+            type: object
+            properties:
+              archiveNumber:
+                type: string
```
liam-lloyd, line 20: `archiveNumber` should be removed here.

packages/api/src/folder/models.ts
liam-lloyd: The `record` table has some columns where `bigint`s are used to represent actual numbers, rather than IDs. Because `bigint`s exceed the maximum size of 32-bit integers, `tinypg` maps them to strings, which we want for IDs, but not for actual numbers. `ArchiveRecordRow` exists to represent what we get out of the database, which we then convert to the actual data model of `ArchiveRecord`. In the case where we can get data matching the data model directly from the database, I don't think we need a separate interface for the database output (we can always add one if the situation changes).

packages/api/src/folder/service.ts
liam-lloyd: `FolderRow` has some required properties that this won't, right? I think because we're returning such a small object, and we're not using that object outside this function, I'd just do
`.query<{folderId: string}>(query, columnsForUpdate)`

packages/api/src/folder/validators.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export function validatePatchFolderRequest(
+  data: unknown
+): asserts data is PatchFolderRequest {
+  const validation = Joi.object()
+    .keys({
+      ...fieldsFromUserAuthentication,
+      archiveNumber: Joi.string().optional().allow(null),
+      archiveId: Joi.number().integer().optional().allow(null),
```
liam-lloyd, line 25: `archiveNumber` and `archiveId` should be removed here.

packages/api/src/folder/service.ts
```diff
@@ -0,0 +1,69 @@
...
+    UPDATE folder
+    SET ${updates.join(", ")}
+    WHERE folderid = :folderId
+    RETURNING folder.folderid AS "folderId"
+  `;
+
+  return query.trim();
+}
+
+export const patchFolder = async (
```
liam-lloyd, line 51: This function needs a mechanism to ensure the authenticated user is updating a folder they have access to

liam-lloyd: And we missed this on PATCH /record, so here's a ticket to fix that https://permanent.atlassian.net/browse/PER-9942

packages/api/src/folder/controller.test.ts
```diff
@@ -0,0 +1,177 @@
...
+  test("expect to log error and return 500 if database update is ok but the database select fails", async () => {
+    const testError = new Error("test error");
+    const originalBehavior = db.sql.bind(db);
+    jest.spyOn(db, "sql").mockImplementation((name, ...params) => {
+      if (name === 'folder.queries.get_folder_by_id') {
+        throw testError;
+      }
+
+      return originalBehavior(name, params)
+    });
```
liam-lloyd, line 145: `mockImplementationOnce()` is what I usually use for these cases where I need the first DB call to succeed an a later one to fail: https://github.com/PermanentOrg/stela/blob/8130f222a325214ffef2f6af1914a138f059c46a/packages/api/src/billing/controller.test.ts#L615

packages/api/src/folder/service.ts
```diff
@@ -0,0 +1,96 @@
...
+    UPDATE folder
+    SET ${updates.join(", ")}
+    WHERE folderid = :folderId
+    RETURNING folder.folderid AS "folderId"
+  `;
+
+  return query.trim();
+}
+
+const canPatchFolder = async (
```
liam-lloyd, line 50: We also need to check the role of the account in the relevant archive (https://permanent.zohodesk.com/portal/en/kb/articles/roles-for-collaboration-and-sharing), and it's possible that it has access via an entry in the `access` table (which represents shares) rather than the `account_archive` table.

liam-lloyd, line 50: Also, if this errors in any case where they don't have the necessary permission, I'd call it `validateCanPatchFolder` and return `Promise<void>`

packages/api/src/folder/service.ts
```diff
@@ -0,0 +1,96 @@
...
+    "account.queries.get_account_archive",
+    {
+      archiveId: folder.archiveId,
+      email: emailFromAuthToken,
+    }
+  );
+  const accountArchive = accountArchiveResult.rows[0];
+
+  if (!accountArchive) {
+    throw new createError.Unauthorized(
```
liam-lloyd, line 67: This should be `Forbidden` instead of `Unauthorized`

packages/api/src/folder/controller.test.ts
iulianvsp: I did that, but I'm seeing some issues with this approach.

1. I would like to mock just one sql call, I want the .sql call to actually reach the database because I want to check if the update actually happened in the DB. I did a .sql call inside the test to check that, but it will just return my mocked response.

2. This is highly dependent on the query order sequence. Adding/Removing one sql call at some point in our code may break a lot of tests and we would have to redo the mocks based on the new order of the sql queries.

Isn't there a way to mock the response based on the function params?

liam-lloyd: Though I don't know if that will provide the behavior you're looking for of "return a mocked response for this set of arguments, otherwise behave as normal". I think it would be reasonable, within this test, to check that the correct update call to db.sql is made, rather than checking that the database itself is updated. Other tests ensure that the update call correctly issued will result in an update.

iulianvsp: Pushing the changes without jest-when, but looking into how can we use it. I think it's the more sustainable option.

packages/api/src/folder/service.ts
iulianvsp: Ok, I'm not 100% of my solution to this. I see that folders can have multiple folder links and the access info is based on folder link id not on folder id.

packages/api/src/folder/service.ts
iulianvsp: After we validate the access check for folders, I'll pick up this one

packages/api/src/folder/service.ts
```diff
@@ -0,0 +1,124 @@
...
+    })
+    .catch((err) => {
+      logger.error(err);
+      throw new createError.InternalServerError(
+        "Failed to retrieve account archive"
+      );
+    });
+  const accountArchive = accountArchiveResult.rows[0];
+  if (!accountArchive) {
+    // ??? How do I make sure I get the correct folder link id?
```
liam-lloyd, line 94: You are free to assume that folders will only have one non-deleted folder link. It hasn't been possible to create a folder or record with multiple folder links since ~2020. There are a relatively small number of folders and records with multiple folder links in the database, but we've been treating these as known broken data for as long as I've been at Permanent.

packages/api/src/folder/service.ts
```diff
@@ -0,0 +1,124 @@
...
+    UPDATE folder
+    SET ${updates.join(", ")}
+    WHERE folderid = :folderId
+    RETURNING folder.folderid AS "folderId"
+  `;
+
+  return query.trim();
+}
+
+const validateCanPatchFolder = async (
```
liam-lloyd, line 74: I'd consider combining the three queries this function uses into a single `get_folder_permissions_data`. Maybe that means this whole function moves to the access folder and that becomes general home for access checks like this.

packages/api/src/folder/service.ts
```diff
@@ -0,0 +1,122 @@
...
+      archiveId: folder.archiveId,
+      email: emailFromAuthToken,
+    })
+    .catch((err) => {
+      logger.error(err);
+      throw new createError.InternalServerError(
+        "Failed to retrieve account archive"
+      );
+    });
+  const accountArchive = accountArchiveResult.rows[0];
```
liam-lloyd, line 92: If the permissions come by way of an `account_archive` record, we need to check that it has edit-level permissions. Note that the Manager role exists on archive memberships, but does not exist on shares (and therefore does not exist on `access` records)

packages/api/src/folder/controller.test.ts
```diff
@@ -0,0 +1,203 @@
...
+  await db.sql("fixtures.create_test_folders");
+};
+
+const clearDatabase = async (): Promise<void> => {
+  await db.query(
+    "TRUNCATE event, account_archive, account, archive, folder CASCADE"
+  );
+};
+
+describe("patch folder", () => {
```
liam-lloyd, line 29: I think it would be worth adding a test where the edit permission comes from the `access` table rather than `account_archive`

packages/api/src/access/permission.ts
```diff
@@ -0,0 +1,43 @@
...
+const checkCanEditAccountArchive = async (
+  archive: GetAccountArchiveResult
+): Promise<boolean> => {
+  let canEdit = false;
+  if (
+    [
+      AccessRole.Admin.toString(),
+      AccessRole.Curator.toString(),
+      AccessRole.Editor.toString(),
+      AccessRole.Owner.toString(),
```
liam-lloyd, line 32: There's a Manager role that should be included here.

liam-lloyd (commented):
Is the empty `/access/index.ts` file necessary?

packages/api/src/account/service.ts
```diff
@@ -125,8 +126,50 @@ const leaveArchive = async ({
...
+      throw new createError.InternalServerError(
+        "Failed to retrieve current account archive"
+      );
+    });
+  if (!currentAccountArchiveResult.rows[0]) {
+    throw new createError.InternalServerError(
+      "Current account archive not found"
+    );
+  }
+  return currentAccountArchiveResult.rows[0];
```
liam-lloyd, line 166: The query here could be returning multiple rows, and all of them matter to us, since we want to check whether the folder is shared with _any_ archive the caller is a member of.

packages/api/src/folder/service.ts
```diff
@@ -0,0 +1,106 @@
...
+    emailFromAuthToken
+  );
+  if (
+    accountArchive &&
+    (await permission.checkCanEditAccountArchive(accountArchive))
+  ) {
+    canEdit = true;
+  }
+  if (!canEdit) {
+    const currentAccountArchive = await accountService.getCurrentAccountArchive(
```
liam-lloyd, line 69: Remember that we also need to check the access role on the `account_archive` for the archive that received a share; if the user only has viewer permissions on the archive they shouldn't be able to edit a share with that archive regardless of the permissions of the share itself.

packages/api/src/folder/service.ts
```diff
@@ -0,0 +1,114 @@
...
+  if (accountArchive && permission.checkCanEditAccountArchive(accountArchive)) {
+    canEdit = true;
+  }
+  if (!canEdit) {
+    const archiveMemberships =
+      await accountService.getCurrentAccountArchiveMemberships(
+        emailFromAuthToken
+      );
+    const editMemberships = await Promise.all(
+      archiveMemberships.map(async (archiveMembership) => {
```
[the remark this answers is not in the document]
iulianvsp: I tried, but no, filter does not work with an async callback

packages/api/src/folder/controller.test.ts
```diff
@@ -0,0 +1,311 @@
...
+      ) => {
+        req.body.emailFromAuthToken = "test+1@permanent.org";
+        req.body.userSubjectFromAuthToken =
+          "b5461dc2-1eb0-450e-b710-fef7b2cafe1e";
+        next();
+      }
+    );
+
+    jest
+      .spyOn(folderAccess, "getAccessByFolder")
```
liam-lloyd, line 140: We should either not mock this, or add unit tests for access.ts, because it is currently not covered by tests.

packages/api/src/folder/service.ts
```diff
@@ -0,0 +1,114 @@
...
+            return archiveMembership;
+          }
+        }
+        return null;
+      })
+    );
+    if (editMemberships.find((membership) => membership !== null)) {
+      canEdit = true;
+    }
+  }
```
liam-lloyd, line 87: I realized in testing that there should be two possible error messages here: the existing 403 if the user is a Viewer or Contributor to the archive where the folder lives, or if they have access to a Viewer or Contributor share of the file, and a 404 in any other case where they can't access it (that is to say, if the user isn't allowed to see the file, we shouldn't confirm for them that it exists).

liam-lloyd (commented):
Functionally, this looks good and is working. There's one hole in the test coverage, and I realized that some cases where this returns 403s should actually return 404s.

packages/api/src/folder/controller.test.ts
iulianvsp: Was able to not mock it in some tests. Others still need mocking to to cover other scenarios unavalilable with the curent fixtures. I would say changing fixtures to not mock this at all will be a bit difficult.

How can we see the test code coverage?

packages/api/src/folder/service.ts
iulianvsp: I see, updated the code.

packages/api/src/folder/controller.test.ts
liam-lloyd: You can add the `--coverage` flag to the test command in `packages/api/package.json`. The cases we're missing right now are the case where this is an `access` entry for the folder, and the case where the database call fails.

packages/api/src/folder/service.ts
liam-lloyd: I think this covers every case but one - if the caller has view permissions on an archive that has an editable share of the folder, they should get a 403 instead of a 404.

liam-lloyd (approved):

PR #151, opened 2024-12-03
PER-9935 Support V2 requests to /record/copy

iulianvsp:


packages/api/src/record/validators.ts
```diff
@@ -42,3 +42,28 @@ export function validatePatchRecordRequest(
...
   }
 }
+
+export function validateCopyRecordRequest(
+  data: unknown
+): asserts data is CopyRecordRequest {
+  const validation = Joi.object()
+    .keys({
+      ...fieldsFromUserAuthentication,
+      RecordVO: {
```
liam-lloyd, line 52: The concept of a VO isn't used in `stela`

packages/api/src/record/validators.ts
```diff
@@ -42,3 +42,28 @@ export function validatePatchRecordRequest(
...
+        recordId: Joi.number().integer().required(),
+        archiveNbr: Joi.string().optional().allow(null),
+        folder_linkId: Joi.number().integer().optional().allow(null),
+        parentFolder_linkId: Joi.number().integer().optional().allow(null),
+        parentFolderId: Joi.number().integer().optional().allow(null),
+        folderId: Joi.number().integer().optional().allow(null),
+        uploadFileName: Joi.string().optional().allow(null),
```
liam-lloyd, line 59: I don't think we need any of these fields

packages/api/src/record/controller.ts
```diff
@@ -64,3 +66,21 @@ recordController.patch(
     }
   }
 );
+
+recordController.post(
+  "/copy",
```
liam-lloyd, line 71: I think the path here should be "/{recordId}/copy"

packages/api/src/record/validators.ts
```diff
@@ -42,3 +42,28 @@ export function validatePatchRecordRequest(
...
 }
+
+export function validateCopyRecordRequest(
+  data: unknown
+): asserts data is CopyRecordRequest {
+  const validation = Joi.object()
+    .keys({
+      ...fieldsFromUserAuthentication,
+      RecordVO: {
+        recordId: Joi.number().integer().required(),
```
liam-lloyd, line 53: `recordId` should probably be a URL parameter for this endpoint

packages/api/src/record/validators.ts
```diff
@@ -42,3 +42,28 @@ export function validatePatchRecordRequest(
...
+        archiveNbr: Joi.string().optional().allow(null),
+        folder_linkId: Joi.number().integer().optional().allow(null),
+        parentFolder_linkId: Joi.number().integer().optional().allow(null),
+        parentFolderId: Joi.number().integer().optional().allow(null),
+        folderId: Joi.number().integer().optional().allow(null),
+        uploadFileName: Joi.string().optional().allow(null),
+      },
+      FolderDestVO: { folder_linkId: Joi.number().integer().required() },
```
liam-lloyd, line 61: I'd use `folderId` rather than `folder_linkId`. Ideally callers of the API never need know what a `folder_link` is.

packages/api/src/record/validators.ts
```diff
@@ -42,3 +42,28 @@ export function validatePatchRecordRequest(
...
+      archiveId: Joi.number().integer().optional().allow(null),
```
liam-lloyd, line 62: I don't think we need this field?

packages/api/src/record/service/copy.ts
```diff
@@ -0,0 +1,81 @@
...
+    folderLinkId: recordData.FolderDestVO.folder_linkId.toString(),
+  });
+
+  const record = await getRecordFromRecordVo(recordData.RecordVO);
+  if (!record) {
+    throw new InternalServerError("RecordVO does not exist");
+  }
+  // Do we need need to check for read access?
+
+  // Do we need to check for create permission in target folder?
```
liam-lloyd, line 42: Yes. The relevant places to check are the `account_archive` entries binding the calling account to the archive(s) where the record and destination folder live, plus any `access` records associated with the record and/or folder

packages/api/src/record/validators.ts
liam-lloyd, line 62: Also note that IDs are always strings in `stela`. Numeric types should only be used if we expect to do arithmetic or numeric comparison with a value

packages/api/src/record/service/copy.ts
```diff
@@ -0,0 +1,81 @@
...
+  if (record === undefined) {
+    throw new InternalServerError("RecordVO does not exist");
+  }
+  return record;
+};
+
+export const copyRecord = async (
+  recordData: CopyRecordRequest
+): Promise<FolderRow> => {
+  // TODO check space over - How to check if there is enough space?
```
liam-lloyd, line 31: The `account_space` table keeps the running total of the space available to an account. The account to check is the calling account, unless the archive where the destination folder lives has `payerAccountId` set, in which case that's the account to check.

packages/api/src/record/service/copy.ts
liam-lloyd: Also we should check for publish permissions on the record if it's being copied to a public folder

liam-lloyd: Also checking for write permissions on the destination should include the check that it's not a connector folder.

liam-lloyd (commented):
In investigating the existing copy endpoint, I realized that while this is certainly a useful effort, it isn't actually pressing, the capability that I thought we needed a V2 version for actually exists already. So I think we should still do this, but you're welcome to prioritize other tickets over it.

packages/api/src/record/validators.ts
iulianvsp: Ah, so we don't want to keep backwards compatibility then? I was trying to keep the request format the same.

liam-lloyd: Yeah, this does not need to be backwards compatible

PR #153, opened 2024-12-07
Per 9878 add share api designs

liam-lloyd:
The view the generated documentation site, run 
```redocly build-docs packages/api/docs/future/api.yaml```
and open the HTML it creates

@meisekimiu I'm particularly interested in your input on whether there are any missing or extraneous fields in the folder object.


liam-lloyd:
For reference, the Share/Publish design these endpoints seek to enable is described here: https://docs.google.com/document/d/1QdYmvgjoEypLrozbL33A1v7d7wvAWQcI4zIjNEQQnSQ/edit?tab=t.0#bookmark=id.wud8g2z4xt3k
Most of this functionality can be built with existing endpoints, this commit describes the new ones we'd need to build.

meisekimiu (approved):

PR #154, opened 2024-12-12
Add endpoints for third party developers to future API design

liam-lloyd:
We want to create a smooth process for third-party developers to integrate their apps with Permanent, or build their entire apps on top of Permanent. In pursuit of this, this commit adds several endpoints to our future API design, covering the creation of third-party developer accounts, the creation and management of third party applications in FusionAuth, and login via OAuth.

### DIFF
--- a/packages/api/docs/future/models/application.yaml
+++ b/packages/api/docs/future/models/application.yaml
@@ -1,0 +1,27 @@
+applicationScope:
+  type: string
+  enum:
+    - record.create
+    - folder.view
+    - folder.create
+    - archive.view
+    - archive.create
+    - storage.view
+
+application:
+  type: object
+  required: [id, secret, scopes, redirectUris]
+  properties:
+    id:
+      type: string
+    secret:
+      type: string
+    scopes:
+      type: array
+      items:
+        type: string
+    redirectUris:
+      type: array
+      items:
+        type: string
+        format: uri
--- a/packages/api/docs/future/paths/application.yaml
+++ b/packages/api/docs/future/paths/application.yaml
@@ -1,0 +1,152 @@
+applications:
+  post:
+    summary: >
+      Register a new application to use the Permanent API.
+      Authorization token must belong to a registered developer account.
+    operationId: createApplication
+    security:
+      - bearerHttpAuthentication: []
+    tags:
+      - application
+    requestBody:
+      content:
+        application/json:
+          schema:
+            type: object
+            required: [scopes]
+            properties:
+              redirectUris:
+                type: array
+                items:
+                  type: string
+                  format: uri
+              scopes:
+                type: array
+                items:
+                  $ref: "../models/application.yaml#/applicationScope"
+    responses:
+      "201":
+        description: The application created
+        content:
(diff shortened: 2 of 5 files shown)


packages/api/docs/future/models/application.yaml
```diff
@@ -0,0 +1,17 @@
+application:
+  type: object
+  required: [id, claims]
```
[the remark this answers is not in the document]
slifty: I wonder if `claims` is supposed to generally be `scopes`

packages/api/docs/future/paths/application.yaml
```diff
@@ -0,0 +1,166 @@
...
+        $ref: "../../shared/errors.yaml#/401"
+      "403":
+        $ref: "../../shared/errors.yaml#/403"
+      "404":
+        $ref: "../../shared/errors.yaml#/404"
+      "500":
+        $ref: "../../shared/errors.yaml#/500"
+  patch:
+    summary: >
+      Update an application's claims.
```
slifty, line 92: This lets someone update more than just the claims / scopes, right?  Might be worth updating the summary to be more generic.

packages/api/docs/future/paths/application.yaml
```diff
@@ -0,0 +1,166 @@
...
+                type: array
+                items:
+                  type: string
+                  format: uri
+              scopes:
+                type: array
+                items:
+                  type: string
+                  enum:
+                    - record.create
```
slifty, line 28: This set of enum values exists in two places (and I think the list is coupled across both, so if one changes the other should too).  I don't think the enum value list can be reused, but the entire enum / type can become a schema definition.

Let's define the items in terms of a referenced schema type of `ApplicationScope` (or whatever name you think is best)

slifty (changes requested):
Looks great.

There's one thing to DRY up, then there's the `claims` vs `scopes` question that @meisekimiu raised

packages/api/docs/future/models/application.yaml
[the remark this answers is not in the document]
liam-lloyd: Yeah, I changed the name and missed it here.

packages/api/docs/future/paths/application.yaml
```diff
@@ -0,0 +1,152 @@
...
+    security:
+      - bearerHttpAuthentication: []
+    tags:
+      - application
+    requestBody:
+      content:
+        application/json:
+          schema:
+            type: object
+            required: [claims]
```
meisekimiu, line 16:
```suggestion
            required: [scopes]
```

packages/api/docs/future/paths/application.yaml
```diff
@@ -0,0 +1,152 @@
...
```
meisekimiu, line 97:
```suggestion
            required: [scopes]
```

meisekimiu (commented):
Should these also be `scopes` instead of `claims`?

slifty (approved):
Thanks for making those changes; this is lookin' clean and correct!

PR #155, opened 2024-12-17
PER-9942 /record/patch could be used to update records the users does…

iulianvsp:
…n't have permission to update


packages/api/src/record/service.ts
```diff
@@ -32,6 +35,55 @@ export const getRecordById = async (requestQuery: {
...
+  }
+  if (!canEdit) {
+    const archiveMemberships =
+      await accountService.getCurrentAccountArchiveMemberships(
+        emailFromAuthToken
+      );
+    const editMemberships = await Promise.all(
+      archiveMemberships.map(async (archiveMembership) => {
+        if (permission.checkCanEditAccountArchive(archiveMembership)) {
+          const accessList = await folderAccess.getAccessByFolder(
```
liam-lloyd, line 65: `access` entries are created for every shared item, including the descendants of shared folders, so you should check for an `access` entry for the record itself rather than its parent (the record could have been shared on its own, rather than as part of a folder).

PR #156, opened 2024-12-18
PER-9942 /record/patch could be used to update records the users does…

iulianvsp:
…n't have permission to update


liam-lloyd (commented):
This looks good, except for the test coverage and error type issue that the PATCH /folder PR has, which are both replicated here.

PR #164, opened 2025-01-11
Trigger terraform validation on PR

liam-lloyd:
Currently, terraform validation Github actions trigger only when changes to the terraform code are merge to main. We'd like to know about validation errors before this, so this commit updates the actions to also run when we create PRs changing the terraform code.


liam-lloyd:
You can test this by committing a change to a terraform file to this branch, seeing that the validation runs, then removing that commit.

meisekimiu (approved):

PR #180, opened 2025-01-28
Support acccess to records via unlisted share token

liam-lloyd:
We want to add unlisted share links, so users can share items in their archive with people who don't have Permanent accounts, without actually publishing those items. To that end, this commit updates the GET /records endpoint to accept the X-Permanent-Share-Token header, and to expose records if the token in that header corresponds to an unlisted share exposing those records. This could be a share of a record itself, or of a record's ancestor folder.


meisekimiu (commented):
Looks good! Will approve once the checks for expired tokens is added!

meisekimiu (approved):

PR #182, opened 2025-02-01
Add PATCH /share-links endpoint

liam-lloyd:
We need to update our share link endpoints to support unlisted links, and this is a good opportunity to migrate these endpoints to stela. To that end, this commit adds a PATCH endpoint for share links which supports unlisted links.

### DIFF
--- a/packages/api/src/share_link/controller.test.ts
+++ b/packages/api/src/share_link/controller.test.ts
@@ -19,6 +19,7 @@ const loadFixtures = async (): Promise<void> => {
   await db.sql("share_link.fixtures.create_test_records");
   await db.sql("share_link.fixtures.create_test_folders");
   await db.sql("share_link.fixtures.create_test_folder_links");
+  await db.sql("share_link.fixtures.create_test_shareby_urls");
 };
 
 const clearDatabase = async (): Promise<void> => {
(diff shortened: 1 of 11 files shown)


packages/api/src/share_link/controller.test.ts
```diff
@@ -331,3 +332,217 @@ describe("POST /share-links", () => {
...
+
+    const shareLink = await retrieveShareLink("6");
+    expect(shareLink?.maxuses).not.toBeNull();
+    expect(shareLink?.expiresdt).not.toBeNull();
+    expect(shareLink?.defaultaccessrole).toEqual("access.role.editor");
+    expect(shareLink?.unrestricted).toEqual(false);
+    expect(shareLink?.autoapprovetoggle).toEqual(0);
+  });
+
+  test("should return 400 if the request is empty", async () => {
```
slifty, line 502: Right now this actually sends an empty object (rather than an empty request).  I could see value in testing both, but as named I would expect it to `send()`.

packages/api/src/share_link/controller.test.ts
```diff
@@ -331,3 +332,217 @@ describe("POST /share-links", () => {
...
+    await agent.patch("/api/v2/share-links/1001").send({}).expect(400);
+  });
+
```
slifty, line 505: Can we add a test for covering cases when malformed json is sent?

packages/api/src/share_link/service.ts
```diff
@@ -69,4 +74,141 @@ const createShareLink = async (
   };
 };
 
-export const shareLinkService = { createShareLink };
+const getShareLinkColumnValuesForUpdate = (
+  requestData: UpdateShareLinkRequest
+): UpdateableShareLinkColumns => {
+  const columnsToUpdate: UpdateableShareLinkColumns = {};
+  if (requestData.permissionsLevel !== undefined) {
+    columnsToUpdate.defaultaccessrole = `access.role.${requestData.permissionsLevel}`;
```
slifty, line 82: I'd expect camelCase here `defaultAccessRole` -- or maybe this is just the existing column name / you don't get to decide the name in which case please carry on!

packages/api/src/share_link/service.ts
```diff
@@ -69,4 +74,141 @@ const createShareLink = async (
...
+      (data.accessRestrictions === "none" ||
+        data.accessRestrictions === undefined) &&
+      ((data.permissionsLevel && data.permissionsLevel !== "viewer") ||
+        (data.maxUses !== undefined && data.maxUses !== null))) ||
+    (data.accessRestrictions === "none" &&
+      (shareLinkResult.rows[0].permissionsLevel !== "viewer" ||
+        shareLinkResult.rows[0].maxUses !== null))
+  ) {
+    throw new createError.BadRequest(
+      "Unlisted links cannot have restricted uses or permissions greated than 'viewer'"
```
slifty, line 188: `greated` => `greater`

packages/api/src/share_link/service.ts
```diff
@@ -69,4 +74,141 @@ const createShareLink = async (
...
+      logger.error(err);
+      throw new Error("Failed to get share link");
+    });
+
+  if (shareLinkResult.rows[0] === undefined) {
+    throw new createError.NotFound("Share link not found");
+  }
+
+  if (
+    (shareLinkResult.rows[0].accessRestrictions === "none" &&
```
slifty, line 178: This if block is pretty hard to parse through; would it be possible to break this into a few named helper functions (can be define in this file / right above).  The purpose of the function wouldn't be re-usability necessarily, but rather legibility.

packages/api/src/share_link/service.ts
```diff
@@ -69,4 +74,141 @@ const createShareLink = async (
...
+  const columnsForUpdate = getShareLinkColumnValuesForUpdate(requestData);
+  const updates = Object.entries(columnsForUpdate)
+    .filter(([_, value]) => value !== undefined)
+    .map(([key, _]) => `${key} = :${key}`);
+
+  if (updates.length === 0) {
+    throw new createError.BadRequest("Request cannot be empty");
+  }
+
+  const query = `
```
slifty, line 114: I imagine this is following existing patterns for updates, but I wanted to flag that inline SQL is not ideal for various reasons (for instance, if sql is in a `.sql` file you can lint it).

For PATCH queries you can conditionally update by passing `NULL` / `undefined` for values you aren't trying to update and then leverage COALESCE() to fall back on the column's current value in cases where no update should happen.

Then for cases when you want to explicitly PATCH a column to a `NULL` value you can instead use a flag like `:setFooToNull` and then `CASE / WHERE` on that flag to explicitly update that column to `NULL`.

This is not intended as a full blown refactor request, but rather just something for you to consider as it might simplify the PR overall and be a cleaner pattern for PATCH updates going forward.

slifty (changes requested):
Typo + a legibility request + one more test that may be worth adding.

I also added a thought about an alternative approach to the query itself which may be cleaner and fit in with the use of `.sql` files for queries instead of inline sql.

packages/api/src/share_link/service.ts
liam-lloyd: This is about matching the column name, but I think the need for that will dissipate if I handle the update query as you suggest below.

packages/api/src/share_link/service.ts
liam-lloyd: Oh yeah, this is a much better approach. The `CASE / WHERE` solution didn't occur to me, which is why I didn't do this in the first place.

PR #186, opened 2025-03-03
Get folders without folder_size from GET /folder

liam-lloyd:
Folder size is calculated asynchronously, which means there are times folders exist without a corresponding folder_size. This commit updates GET /folder to return such folders (with size set to null).

### DIFF
--- a/packages/api/docs/future/models/folder.yaml
+++ b/packages/api/docs/future/models/folder.yaml
@@ -16,7 +16,6 @@ folder:
   type: object
   required:
     - id
-    - size
     - displayName
     - archive
     - status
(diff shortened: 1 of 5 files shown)


slifty (approved):
It doesn't look like this would have an impact on the sdk but you are in a better position to know for sure (are we not using the SDK for stela calls by the frontend?)

liam-lloyd:
> It doesn't look like this would have an impact on the sdk but you are in a better position to know for sure (are we not using the SDK for stela calls by the frontend?)

So far the SDK is not covering any stela endpoints (and I don't think any client outside of SFTP is using the SDK)

PR #187, opened 2025-03-04
Per 10003 add get share links

liam-lloyd:


packages/api/docs/future/paths/share_link.yaml
```diff
@@ -65,15 +65,24 @@ share-link:
...
-      required: true
-      description: A share link's ID, or token
-      schema:
-        type: string
   get:
+    parameters:
+      - name: shareTokens
+        in: query
+        required: false
+        description: An array share links' tokens. Required if shareLinkIds is not set.
```
slifty, line 73: minor typo `array *of* share links'`

packages/api/docs/future/paths/share_link.yaml
```diff
@@ -65,15 +65,24 @@ share-link:
...
+        schema:
+          type: array
+          items:
+            type: string
+      - name: shareLinkIds
+        in: query
+        required: false
+        description: An array share links' ids. Required if shareTokens is not set.
```
slifty, line 81: Typo here as well.

packages/api/src/share_link/service.ts
```diff
@@ -69,4 +73,120 @@ const createShareLink = async (
...
+  currentShareLink: ShareLink
+): boolean =>
+  updateRequest.accessRestrictions === "none" ||
+  (currentShareLink.accessRestrictions === "none" &&
+    updateRequest.accessRestrictions === undefined);
+
+const postUpdatePermissionsLevel = (
+  updateRequest: UpdateShareLinkRequest,
+  currentShareLink: ShareLink
+): "contributor" | "editor" | "manager" | "owner" | "viewer" =>
```
slifty, line 87: You know me -- I think it's clearer to have this kinda thing be in a named alias (or an enum) jfrom a separation of code responsibilities PoV (e.g. this line should just be saying what type of thing is returned; I wouldn't expect the postUpdatePermissionsLevel function definition to also be the place where we define the complete set of PermissionLevels)

Side note: if the linter doesn't like a thing that you want to be able to do (union of strings) I for one am OK with changing it to allow!

packages/api/src/share_link/validators.ts
```diff
@@ -49,3 +49,59 @@ export function validateCreateShareLinkRequest(
...
+): asserts data is UpdateShareLinkRequest {
+  const validation = Joi.object()
+    .keys({
+      ...fieldsFromUserAuthentication,
+      permissionsLevel: Joi.string()
+        .when("accessRestrictions", {
+          is: "none",
+          then: Joi.valid("viewer"),
+          otherwise: Joi.valid(
+            "viewer",
```
slifty, line 64: FWIW if you do go the enum route you would be able to do something like `Object.values(PermissionTypes)` here and avoid the repetition.

packages/api/src/share_link/service.ts
liam-lloyd: As it happens, I know that an updated version of the linter switches its defaults to allow aliases for these. I think, as a separate PR, I'll make that update, and then standardize us on aliased unions for these sorts of things.

PR #188, opened 2025-03-04
Update Github Action runners to Ubuntu 24.04

liam-lloyd:
Because they're retiring the 20.04 runners that we're using now.


slifty (approved):
Looks good!  I did a local search for `ubuntu-` and did not see any missed instances.

PR #198, opened 2025-03-11
Add createdAt and updatedAt fields to file objects

liam-lloyd:
The SFTP service used the createdAt and updatedAt fields of file objects; thus they should be included in file objects returned by this API.


slifty (approved):
Looks great and thank you for adding this!

PR #200, opened 2025-03-13
Support authentication via SFTP token for GET /folder endpoints

liam-lloyd:
The SFTP service is it's own application in Fusionauth, and SFTP users should be able to access the GET /folder endpoints offered by this API. To enable this, this commit updates the auth middleware used by those endpoints to check tokens against both the backend and sftp applications, and succeed if the token comes from either.

Note that this is not a good or scalable way to do this. In the short term, we should rewrite this to validate JWTs on our own without making calls to FusionAuth, and then check the aud field of the JWTs against an internal list of application IDs to see what kind of access the caller should have. In the longer term, this check should be based on claims assigned to applications, rather than on applications directly.


slifty (approved):
This is an elegant short term solution for an awkward situation!  Appreciate the commit / PR note about next steps; we should capture those thoughts in an issue.

PR #206, opened 2025-04-02
Updates the account space lambda to use file.size over record.size

liam-lloyd:
Currently, the account space update lambda uses record.size to determine the size of the record. However, at the point that the lambda runs, that value is set based on a parameter provided by the user, while file.size is set based on the actual file size as reported by S3. Thus, we should use file.size instead.


slifty (approved):
Nice catch

PR #208, opened 2025-04-04
Add /archive/:archiveId/backfill-ledger admin endpoint

liam-lloyd:
We have some archives in the system where storage wasn't historically tracked via the application, but rather by hand, in order to facilitate free storage for nonprofit partners. We can now achieve this while still tracking storage in the application, and this in fact preferable for partners because it allows them to track their storage balances more easily. This commit adds an admin tool that can help migrate archives from the old accounting system to the new one by backfilling the storage ledger for all items in the archive that weren't originally included therein.

### DIFF
--- a/packages/api/src/archive/controller/backfill_ledger.test.ts
+++ b/packages/api/src/archive/controller/backfill_ledger.test.ts
@@ -1,0 +1,116 @@
+import request from "supertest";
+import type { Request, NextFunction } from "express";
+import createError from "http-errors";
+import { logger } from "@stela/logger";
+import { app } from "../../app";
+import { verifyAdminAuthentication } from "../../middleware";
+import { db } from "../../database";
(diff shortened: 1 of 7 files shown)


slifty (commented):
This looks good but before I formally approve I wanted to raise a thought:

If this endpoint exists to correct a legacy data issue, I think it would be cleaner to address that legacy data issue in a single act wherever the database schema is formed and managed; specifically I think this should be a migration in the PHP code base that fixes all of the data in one swoom, which also makes whatever schema changes are appropriate to remove the "old" stuff.

Let's briefly talk about this after sprint planning.

slifty (approved):
We talked today about the "API vs migration" question -- API is the way to do it for now, particularly because we want to be able to coordinate with partners as we update their state.

PR #209, opened 2025-04-04
Update EKS clusters to Kubernetes 1.32

liam-lloyd:
Kubernetes 1.30 will exit standard support on EKS in July (and thus become more expensive). To avoid cost increases, this commit updates our EKS clusters to the lastest Kubernetes version, 1.32.


slifty (approved):
This looks good -- do we have any written "policies" around infra versioning in general?  Kubernetes cluster versions, node versions, OS versions, etc.

It would be good to document that and then put in some cyclical practices to adhere to them (e.g. first week of each month we upgrade { whatever is supposed to be upgraded })

Issue #196, opened 2025-03-11
Consider serving a swagger page


slifty:
The readme references use of ReDoc to build api documentation, but it would be neat if the project just rendered its documentation of the box without requiring additional setup.

[swagger UI Express](https://www.npmjs.com/package/swagger-ui-express) would allow us to render a swagger page at `.../api` (or whatever path we want).  This is extra handy since it provides a functional UX for calling endpoints and therefore allows manual testing.

If we decide to implement this, we could use [the PDC service](https://github.com/PhilanthropyDataCommons/service/) as an example project that follows a similar pattern.

liam-lloyd:
Update: we're about to launch a redoc page via Github Pages; we should in the future look at using redoc-express or swagger-ui-express to serve this page from the same server as the stela API, and without having to commit the generated HTML

PR #210, opened 2025-04-12
Reorganize docs and commit docs site

liam-lloyd:
We would like to make our API documentation available via a website hosted on Github Pages. To that end, this commit adds the site output by redocly to the repo. Also, to make the docs more useful, it unifies the "present" and "future" docs into a single body, noting at the endpoint level which are not yet implemented.


liam-lloyd:
Added an unimplemented tag, but also kept the parentheticals because they're very prominently placed and I want it to be quite clear with endpoints don't exist yet (and because I didn't want to remove all other tags from the unimplemented endpoints to force them to only appear in the unimplemented section)

liam-lloyd:
Note that this is an intentionally quick and imperfect way of getting a docs page live; see https://github.com/PermanentOrg/stela/issues/196 for discussion of where we want to go with this next.

PR #211, opened 2025-04-14
Fix folder tests

liam-lloyd:
We recently made folder.locnid a foreign key. This means the folder tests need to load the locn fixtures before the folder fixtures now.


liam-lloyd:
I tried to test for this before I put up the folder.locnid PR, but evidently my environment was not in the right state to catch it when I did.

PR #219, opened 2025-04-21
Switch to EditorConfig

### LINKED ISSUE #217: Switch to using editorconfig
There are .vscode editor settings, but using `.editorconfig` would be a more generalized IDE configuration tool (VSCode has an editorconfig plugin).

We use this in other repositories as well so we can copy those settings.

slifty:
This PR (1) sets up an editorconfig file in line with our other repositories and (2) removes developer-specific settings.

@liam-lloyd if you're up for it, you could add `.vs-code` to your global gitignore (`~/.gitignore`) which means it wouldn't be committed with a project unless you explicitly wanted to.  Generally I believe this is a best practice to separate individual dev env settings from the general code base.  I'm also happy to leave that part in if that's preferred!

Resolves #217

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-  "editor.tabSize": 2
-}


liam-lloyd:
> This PR (1) sets up an editorconfig file in line with our other repositories and (2) removes developer-specific settings.
> 
> @liam-lloyd if you're up for it, you could add `.vs-code` to your global gitignore (`~/.gitignore`) which means it wouldn't be committed with a project unless you explicitly wanted to. Generally I believe this is a best practice to separate individual dev env settings from the general code base. I'm also happy to leave that part in if that's preferred!
> 
> Resolves #217

I would do this, but I don't use VSCode, so it shouldn't be necessary (it was one of the VSP folks who added the .vs-code deleted here).

liam-lloyd (approved):

PR #273, opened 2025-04-28
 Ignore records uploaded to nonprofit accounts in account space updater

liam-lloyd:
We aren't supposed to charge nonprofit type archives for storage (their storage balances are instead tracked manually). Though we are moving away from this system, the account space updater lambda should abide by it for the moment.


slifty (changes requested):
I would expect there to be a test explicitly titled about making sure nonprofit records don't affect account space updates.

I realize that the way our fixtures are defined outside of the context of specific test names means you've functionally tested that edge case, but because the test isn't explicit it's harder for someone to look at the tests and understand what is / isn't covered.

For example: in future someone might adjust the fixture in a way that causes this particular case to no longer be properly tested by the main test without necessarily realizing it.

(this might be a more meta-comment that warrants reflection on the way we are using generalized fixtures / it might be cleaner-even-if-more-verbose to 1:1 relationships between fixtures / setup logic and specific tests, but that's a bigger conversation!)

PR #281, opened 2025-04-30
Include location in record objects

liam-lloyd:
The web app expects records to include a location, so this commit updates the V2 API to comply with this expectation.


slifty:
Just wanted to flag that... this PR having -32k lines speaks to why having our generated documentation as part of this repo is probably *not* how we want to do it.

This might probably increase the urgency of making a change.

Assuming including that removal is intentional, could you separate the documentation generation commit from the actual API change commit?

PR #283, opened 2025-05-02
Remove unnecessary body-parser dependency

### LINKED ISSUE #284: Removing body-parser
We use the `body-parser` express middleware which in theory we might not need to now that express supports `.json()` natively.

The hangup is that our authentication system currently decorates `.body` with some auth context information.  I spoke with @liam-lloyd and he agreed that we can move that to another attribute such as`req.auth`, which is safer anyway since `body` is populated by user input directly whereas something like `req.auth`.

So this issue would involve shifting that logic to use `.auth` and then removing the body parser middleware!

slifty:
This PR removes an unnecessary dependency (Express provides a native version of `.json()` middleware now)

Resolves #284

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -100,8 +100,7 @@
 		},
 		"node_modules/@aws-crypto/crc32c": {
 			"version": "5.2.0",
-			"resolved": "https://registry.npmjs.org/@aws-crypto/crc32c/-/crc32c-5.2.0.tgz",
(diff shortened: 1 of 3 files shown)


cecilia-donnelly:
Did you two decide not to move forward with this?

slifty:
@cecilia-donnelly you know what, I didn't -- I just forgot I had started it after pulling into another priority at the time.

I'll finish this up shortly.  What's missing is the replacement of the middleware (I took out body parser but didn't add back `express.json()`)

slifty:
@liam-lloyd oh brilliant patch thank you!

I'm going to review to approve your portion ;)

cecilia-donnelly:
:tada: :tada: :tada:

Issue #215, opened 2025-04-21
Update Stela to Node 22


slifty:
Node 18 is no longer the active version -- we've updated other things to Node 22 so we should do the same here!

As part of this we can move to using a `.node_version` file for our workflows

PR #292, opened 2025-05-07
Deploy docs to a `gh-pages` branch

### LINKED ISSUE #291: Switch to a `gh-pages` branch for documentation
We use Github pages to serve our api documentation from this repository; this is a valid approach (though we may eventually move to a dynamic rendering from express some day: #196)

That said, right now we commit that compiled documentation in main.  This has a few downsides:

1. It requires a manual step from a developer when making API changes
2. It results in large diff commits in our history every time we make an API change
3. We have a compiled resource in our main branch.

GitHub provides a solution to all of our problems: the `gh-pages` branch

We can set up a workflow which, on main merge, will compile the documentation and push the result to `gh-pages` for us.

slifty:
This PR adds a workflow to publish documentation pages to `gh-pages` every time there is a merge to main.

Resolves #291


.github/workflows/deploy_docs.yml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout main
+        uses: actions/checkout@v3
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "22"
```
slifty, line 19: We should use a `.node-version` file and reference it in our workflows, but I didn't want to cause trouble in this PR

.github/workflows/deploy_docs.yml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout main
+        uses: actions/checkout@v3
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "18"
```
slifty, line 19: Until #215 is done I'm sticking with the version of node other workflows use

slifty:
For one cycle I added the pages push to this PR's workflow, and it did indeed push to gh-pages (https://github.com/PermanentOrg/stela/tree/gh-pages)

So it seems to work!

.github/workflows/test.yml
```diff
@@ -46,3 +46,18 @@ jobs:
...
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "18"
+
+      - name: Install Redoc CLI
+        run: npm install -g @redocly/cli
+
+      - name: Generate HTML from OpenAPI spec
+        run: npx @redocly/cli build-docs ./packages/api/docs/src/api.yaml -o docs/index.html
```
liam-lloyd, line 63: I'm not sure that we need this; I think the `redocly lint` that we run as part of linting will catch anything that would make a build fail (though the docs don't declare that, it's just what I've seen so far).

liam-lloyd (approved):

slifty:
For one cycle I added the pages push to this PR's workflow, and it did indeed push to gh-pages (https://github.com/PermanentOrg/stela/tree/gh-pages)

So it seems to work!

I also updated our settings so that pages is serving from `gh-pages`

.github/workflows/test.yml
slifty: I'll take this out then!

PR #306, opened 2025-05-17
Per 10190 add file url refresh cron

liam-lloyd:


slifty (approved):
Solid change!

Just to document here: I had asked Liam separately about this approach vs a "refresh on demand" approach and he had some good thinkin' regarding decoupling CDN URL generation from content-access API endpoints / having to spread CDN url management across a bunch of places.

It could be that some day there is a way to have our cake and eat it too (e.g. have some architectural component that doesn't run via cron but somehow decorates data requests with CDN refreshes on demand), but that is not something we need to think about today at all, and this does the trick nicely!

PR #315, opened 2025-05-21
Expose 429 errors during optional token processing

liam-lloyd:
When processing tokens for optionally authenticated endpoints, errors are ignored, because authentication isn't necessary. However, 429 errors indicate that the client should back off on calls to this API regardless of whether the requested resource required authentication, so they should be propagated to the caller. This commit updates the authentication middleware to do that.


slifty (approved):
Need to fix failing tests but otherwise this looks good!

PR #328, opened 2025-06-03
Make share links accessible via share token without JWTs

liam-lloyd:
The UI needs to be able to determine the access restrictions of a share link without a logged in user. Thus this commit makes share links accessible without a JWT, provided they're accessed by share token and not by ID.

### DIFF
--- a/packages/api/src/share_link/controller.test.ts
+++ b/packages/api/src/share_link/controller.test.ts
@@ -4,7 +4,10 @@ import createError from "http-errors";
 import { v4 as uuidv4 } from "uuid";
 import { when } from "jest-when";
 import { app } from "../app";
-import { verifyUserAuthentication } from "../middleware";
+import {
+	extractUserEmailFromAuthToken,
+	verifyUserAuthentication,
+} from "../middleware";
 import { db } from "../database";
 import type { ShareLink } from "./models";
 
@@ -46,7 +49,7 @@ const retrieveShareLink = async (
 	folderLinkId: string,
 ): Promise<ShareLinkRow | undefined> => {
 	const shareLinkResult = await db.query<ShareLinkRow>(
-		`SELECT 
+		`SELECT
         status,
         urltoken,
         shareurl,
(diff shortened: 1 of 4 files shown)


packages/api/src/share_link/controller.test.ts
```diff
@@ -611,29 +608,29 @@ describe("GET /share-links", () => {
...
+				).emailFromAuthToken = "not_an_email";
 				next();
 			},
 		);
 		await agent.get("/api/v2/share-links?shareLinkIds[]=1").expect(400);
 	});
 
 	test("should return an empty list if caller asks for share links they can't access", async () => {
-		(verifyUserAuthentication as jest.Mock).mockImplementation(
+		(extractUserEmailFromAuthToken as jest.Mock).mockImplementation(
```
slifty, line 626: I hadn't seen this pattern (type casting a mocked function as mocked) before so looked into it.  The motivation is just that as you know type casts are an override to the static analysis / if we're wrong about the cast then it causes problems.

It looks like the safer alternative to this approach would be to use [`jest.mocked()`](https://jestjs.io/docs/mock-function-api#jestmockedsource-options).

So you could replace this line with:

```
jest.mocked(extractUserEmailFromAuthToken).mockImplementation(...)
```

And you'd get the mocking type without the type cast!

packages/api/src/share_link/service.ts
```diff
@@ -168,9 +168,9 @@ const updateShareLink = async (
 };
 
 const getShareLinks = async (
-	email: string,
 	shareTokens: string[] | undefined,
 	shareLinkIds: string[] | undefined,
+	email?: string,
```
slifty, line 173: I've tended to succumb to the idea that optional parameters (as opposed to required parameters which can be passed `undefined` as an allowed value) are not ideal; in this PR it looks like email *is* always passed, it's just it can be undefined right?  If so, let's consider making the parameter required but allowing `| undefined`.

slifty (approved):
Two suggestions.

One is a pattern change that would impact a LOT of tests if done with full fidelity, and I'm not suggesting that change be made in this PR.

The other is just a parameter consistency thing

PR #364, opened 2025-06-27
Upgrade to Node 22

liam-lloyd:


Dockerfile.account_space_updater
```diff
@@ -1,4 +1,4 @@
-FROM public.ecr.aws/lambda/nodejs:18 as builder
+FROM public.ecr.aws/lambda/nodejs:22 as builder
```
slifty, line 1: Ignore this thought unless it triggers inspiration.

I'm seeing a lot of duplication of this base (and then duplication of the alpine / non-lambda base).  It makes me wonder if there's a way to DRY this so upgrading the base image would happen in one place.

I did snooping and if we did want to do this it seems like it would involve defining and publishing our own base image (via this repository in something like `/docker/base/Dockerfile` and `/docker/lambda-base/Dockerfile`) and then having these all build off of that base image.

Then updating would involve building and publishing a new base image.

Again just food for thought.

slifty (approved):
Approved with the asterisk of two small DRY ideas (the `.node-version` one is a stronger suggestion since we have other repositories using this approach); the second / docker one really just is an idea.

PR #377, opened 2025-07-02
Replace node-fetch with native fetch

liam-lloyd:
Node 22 has a native fetch function, so we no longer need an extra dependency to provide one.


packages/api/src/directive/controller.test.ts
```diff
@@ -195,7 +194,7 @@ describe("POST /directive", () => {
 			.expect(200);
 
 		const directiveResult = await db.query<Directive>(
-			`SELECT 
+			`SELECT
```
slifty, line 197: (We should definitely set up prettier in here and just run it once and for all!)

slifty (approved):
Yay!  Love to see it

PR #386, opened 2025-07-10
Add eslint love

liam-lloyd:
More PRs turning on inactive linting rules will follow


packages/api/src/archive/controller/get_featured.test.ts
```diff
@@ -34,7 +34,9 @@ describe("getFeatured", () => {
 	test("should retrieve featured archives", async () => {
 		const result = await agent.get("/api/v2/archive/featured").expect(200);
 		expect(result.body).toHaveLength(1);
-		const archive = (result.body as FeaturedArchive[])[0];
```
slifty, line 37: I think this whole test block would be much cleaner as:

```
expect(result.body).toMatchObject([{
  archiveId: 3,
  name: "Jay Rando",
  ...
}]);
```

This also prevents the awkward type cast!  (though maybe I'm writing this way too soon and you address this since I suspect that's the "unsafe type assertions in tests" you were talking about)

OK disregard this comment probably!

eslint.config.mjs
```diff
@@ -89,6 +88,7 @@ export default defineConfig([
 			"@typescript-eslint/unbound-method": "off",
 			"jest/unbound-method": "error",
 			"jest/no-focused-tests": "error",
+			"max-lines": "off",
```
slifty, line 88: This isn't a formal request but a note that now that we're using flat config we can add comments when we modify default rules to provide context for future (e.g. the comment here might be something like "test files are allowed to be long because they're structured and need to be comprehensive")

slifty (changes requested):
The only full request here is the removal of the stale config lines (I'll give a super fast approval after that!)

The other thing for your consideration is I do think it's better / clearer / cleaner to use `expect(...).matchObject(...)` as opposed to individual attributes.  This also makes the test more robust because it means our tests detect situations where expected attributes have been added!

PR #498, opened 2025-09-25
Point this project at our new Archivematica instances

liam-lloyd:
We have new kubernetes-hosted Archivematica instances deployed to all our environments now; this commit updates stela to use these instances rather than the old EC2-hosted ones.


cecilia-donnelly (approved):
I'm not sure where to double check the variable strings but this makes sense!

PR #521, opened 2025-10-14
Update sqlfluff to 3.4.2

liam-lloyd:
It has some bug fixes that we would benefit from, and better indentation defaults.

### DIFF
--- a/packages/api/src/share_link/queries/create_share_link.sql
+++ b/packages/api/src/share_link/queries/create_share_link.sql
@@ -1,60 +1,88 @@
-INSERT INTO shareby_url (
-  folder_linkid,
-  status,
-  urltoken,
-  shareurl,
-  maxuses,
-  unrestricted,
-  autoapprovetoggle,
-  defaultaccessrole,
-  expiresdt,
-  byaccountid,
-  byarchiveid,
-  createddt,
-  updateddt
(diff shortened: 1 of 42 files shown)


packages/api/src/share_link/queries/create_share_link.sql
```diff
@@ -1,60 +1,88 @@
...
+  new_share_link."itemId",
+  new_share_link."itemType",
+  new_share_link.token,
+  new_share_link."permissionsLevel",
+  new_share_link."accessRestrictions",
+  new_share_link."maxUses",
+  new_share_link."usesExpended",
+  new_share_link."expirationTimestamp",
+  new_share_link."createdAt",
+  new_share_link."updatedAt",
```
cecilia-donnelly: I don't quite see what's going on here either - was there previously a new_share_link.* that I'm missing?

cecilia-donnelly (commented):
Okay, so it looks like the general trend is toward more explicit references, which is definitely nice for readability! Maybe also a new requirement for order in JOIN statements? It looks like sometimes AS "something with quotes" was accepted and sometimes not, which I don't quite follow.

slifty (approved):
I know @cecilia-donnelly has some open comments but just to add my approval for the record!

PR #528, opened 2025-10-16
Add GET /archives endpoint to support archive search

liam-lloyd:
This commit adds a GET /archives endpoint to the API server. The archive object it returns includes contact information for the archive owner if the caller is a Permanent administrator; this is to support a streamlined workflow for identifying candidates for featured archives in the admin dashboard.

### DIFF
--- a/packages/api/docs/src/paths/archive.yaml
+++ b/packages/api/docs/src/paths/archive.yaml
@@ -1,3 +1,38 @@
+archives:
+  parameters:
+    - name: searchQuery
+      in: query
+      required: true
+      schema:
+        type: string
+  get:
+    summary: Get archives whose names match the search query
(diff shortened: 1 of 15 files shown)


packages/api/docs/src/paths/archive.yaml
```diff
@@ -1,3 +1,38 @@
...
+      in: query
+      required: true
+      schema:
+        type: string
+  get:
+    summary: Get archives whose names match the search query
+    description: |
+      Unauthenticated requests return only public archives.
+      Admin-authenticated requests return public archives with owner contact information.
+      User-authenticate requests return public archives and archives of which the caller is a member.
```
cecilia-donnelly: typo: authenticated

PR #531, opened 2025-10-20
Run Archivematica on copied files

liam-lloyd:
Copied files should have corresponding AIPs and DIPs just like directly uploaded files. Running them through Archivematica is the simplest way to achieve this.


cecilia-donnelly (commented):
I tried this and didn't see the copied record appear in the Archivematica dashboard, unfortunately. I did upload one and see it in the "Transfer" tab successfully.

I don't actually know where these `recordcreate` messages are created, or else I would look there to see what might be going on.

cecilia-donnelly (approved):
I copied a document on dev and saw it go through on the Archivematica dashboard. Thanks @liam-lloyd !

PR #547, opened 2025-10-31
Per 10011 add archivematica metadata lambda

liam-lloyd:


cecilia-donnelly (approved):
Thanks for the walkthrough, @liam-lloyd ! I think there were a couple small things we noticed (logging left in?) that you were going to go back and remove.

PR #588, opened 2025-12-05
Per 10160 add storage adjustment endpoint

liam-lloyd:


cecilia-donnelly (commented):
For this kind of PR, a "how to test" section in the description would be helpful.

What I did:
1. Get an admin auth token via a resource grant call to FusionAuth (I have a helper script for this if future testers / QAers want it).
2. POST to the new endpoint (https://api.permanent.org/v2/account/:id/storage-adjustment) with the storageAdjustment amount in GB. `curl -d '{"storageAmount": "1"}' -H "content-type: application/json" "https://local.permanent.org/api/v2/account/2/storage-adjustments" -X POST  --insecure -H "Authorization: Bearer <token>"`

This worked with positive and negative `storageAmount`s. It allowed me to go negative on the overall storage amount, which we do allow in the app, but I think we limit in the app.

The transaction history in the UI is wonky. This may all be happening on the front end, but right now it looks like positive amounts convert to GB correctly but negative do not. We should fix that either here or in the front end. There needs to be a web-app followup to add a user-friendly string instead of the db `admin_adjustment`, as well.

Image

PR #608, opened 2026-01-21
Trigger Archivematica in the local environment

liam-lloyd:
This commit makes two minor changes to make it possible to trigger Archivematica processing when files are uploaded to a local copy of Permanent. The way this is done uses localstack SNS/SQS to trigger a lambda running locally via SAM, which then activates a local-env-specific workflow in the dev instance of Archivematica. To enable this, this commit:
1. Adds a template describing the trigger Archivematica lambda for use by SAM
2. Adds an AWS_ENDPOINT_URL env var to allow SNS messages to be routed to localstack instead of actual AWS
3. Makes the processing workflow used by Archivematica configurable by environment variable.


cecilia-donnelly:
I assume this one is waiting on the devenv PR?

liam-lloyd:
> I assume this one is waiting on the devenv PR?

No, I just needed to add a test to satisfy codecov

PR #614, opened 2026-01-27
Per 10429 add edtf to folder endpoints

liam-lloyd:


cecilia-donnelly (changes requested):
Tiny changes, but generally this looks good!

PR #622, opened 2026-01-29
Per 10403 add edtf sorting trigger tests

liam-lloyd:


liam-lloyd:
Built atop https://github.com/PermanentOrg/stela/pull/614

PR #638, opened 2026-02-05
Use matrix for build CI

### LINKED ISSUE #636: Use a matrix for build CI
The `build.yml` has a fair amount of duplicated logic for the various packages.  This also means every time we add a package we have to copy all of that.

We can use a matrix to DRY up those steps -- it will still run in parallel, but the file itself will be shorter and adding future packages should just be a matter of expanding the matrix.

slifty:
This PR moves our build workflow to use a job matrix; this ensures steps are consistent and avoids duplicated code.

Resolves #636

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-name: Build deployable images
+name: Build
 on:
   pull_request:
   workflow_call:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -34,178 +56,13 @@ jobs:
...
-  build_am_cleanup:
-    needs:
-      - generate_image_tags
-    runs-on: ubuntu-24.04
-    env:
-      AM_CLEANUP_IMAGE_TAG: ${{ needs.generate_image_tags.outputs.AM_CLEANUP_IMAGE_TAG }}
-    steps:
-      - uses: actions/checkout@v6
-      - name: Build Image
-        run: docker build -t $AM_CLEANUP_IMAGE_TAG -f packages/archivematica_cleanup/Dockerfile .
```
slifty: This is the one docker build that does not accept the AWS_RDS_CERT_BUNDLE build arg -- In the DRY I figured it doesn't... actually matter if it's getting that injected / that injection is done for every job.

If that is incorrect, I can absolutely make sure it's conditional / doesn't happen for this particular job.

PR #643, opened 2026-02-06
Pass image tags from staging deploy to build workflow

liam-lloyd:
To avoid running generate image tags more than once, and to ensure they're generated consistently, the prod_deploy
workflow runs image tag generateion and passes the tags to staging_deploy. This commit updates staging_deploy to also pass them to build, so build doesn't regenerate them itself.


cecilia-donnelly (approved):
Got it, makes sense.

PR #644, opened 2026-02-06
Update documentation for adding new services

liam-lloyd:
Recently we've updated certain Github workflows to pass image tags between each other in order to avoid running image tag generation redundantly. This commit updates the README to note that when creating a new service, its image tag will need to be added to the parameters passed between these workflows.


cecilia-donnelly (approved):
Thank you!

PR #645, opened 2026-02-06
Make API tests run in parallel

liam-lloyd:
The API package has a lot of tests, and currently they run sequentially because they all use the same database. This takes ~21 minutes on my machine. This commit updates the database setup so that each Jest worker gets its own database, allowing us to use Jest's parallelism. This cuts the time it takes to run the API tests down to ~7 minutes on my machine.


slifty (approved):
love love love this.

PR #649, opened 2026-02-10
Allow archive load by access role

### LINKED ISSUE #616: Add endpoint for loading a user's archives
We want to be able to load all archives available to a given user.

slifty:
This PR expands the `GET /archives` endpoint to accept an `accessRole` array, allowing users to filter the returned archives in terms of their access to the archives.

A user could also provide no search string OR access role filters and get a list of all archives they have any form of access to.

Resolves #616

### DIFF
--- a/packages/api/docs/src/api.yaml
+++ b/packages/api/docs/src/api.yaml
@@ -55,12 +55,12 @@ paths:
     $ref: "./paths/folder.yaml#/folders-id-children"
   /folders/{id}/share-links:
     $ref: "./paths/folder.yaml#/folders-id-share-links"
-  /archive/{id}:
-    $ref: "./paths/archive.yaml#/archives~1{id}"
-  /archive/{id}/folders/shared:
-    $ref: "./paths/archive.yaml#/archive-id-folders-shared"
   /archives:
     $ref: "./paths/archive.yaml#/archives"
+  /archives/{id}:
+    $ref: "./paths/archive.yaml#/archives~1{id}"
+  /archives/{id}/folders/shared:
+    $ref: "./paths/archive.yaml#/archives-id-folders-shared"
   /applications:
     $ref: "./paths/application.yaml#/applications"
   /applications/{id}:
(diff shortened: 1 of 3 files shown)


packages/api/docs/src/models/access_role.yaml
```diff
@@ -0,0 +1,10 @@
+accessRole:
+  type: string
+  description: The access role that defines a user's permissions for an archive or item.
+  enum:
+    - access.role.viewer
+    - access.role.contributor
+    - access.role.editor
+    - access.role.curator
+    - access.role.manager
+    - access.role.owner
```
liam-lloyd: At the API level I think we should free ourselves from the structure used in the database and just let these be `Viewer`, `Contributor`, etc.

slifty: @liam-lloyd noted -- one concern we should be sure to think is avoiding a situation where we have two separate enums used by the client for the concept of access role (e.g. one for when writing access role values, and one for filtering by those values e.g. in this case)

Would you want us to change the API enums broadly (e.g. for POST / PUT too) and translating those values to the expected database values later?

liam-lloyd: Yes, across the board I think the API should use and surface these friendlier enum values; ideally the clients wouldn't know about `access.role.viewer` format. We're already doing this with access roles in share links: https://permanentorg.github.io/stela/#tag/share-links/operation/createShareLink (although upon checking that documentation, my suggestion is that these ones also be lowercase).

packages/api/docs/src/models/archive_membership_role.yaml
```diff
@@ -0,0 +1,11 @@
...
+  type: string
+  description: The role that defines a user's permissions as a member of an archive.
+  enum:
+    - viewer
+    - contributor
+    - editor
+    - curator
+    - manager
+    - owner
+    - admin
```
liam-lloyd: The admin role was mistakenly brought over from the PHP code, but it doesn't actually exist, so we should remove it from stela

liam-lloyd (approved):

PR #656, opened 2026-02-17
Set provenance flag to false on docker builds

liam-lloyd:
When the provenance flag is set to true (as it is by default), the
BuildKit version used on the latest Ubuntu release will use features in
constructing the image that aren't supported by the lambda runner. Since
we're the only consumer of our images, it is safe for us to turn
provenance off to avoid this issue.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -66,7 +66,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
       - name: Build Image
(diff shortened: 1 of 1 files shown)


slifty:
@liam-lloyd could you give a little clarification (in the commit message probably) about what this actually is / does.  I get that we need to make the change, but it's worth adding that context for future versions of us and current reviewers :D

PR #660, opened 2026-02-18
Update GET /share-links docs to reflect all auth pathways

liam-lloyd:
Currently, the API documentation for GET /share-links doesn't reflect that an auth token is not required for requests that use the shareToken parameter rather than a shareLinkId. This commit updates the docs to reflect that reality.

### DIFF
--- a/packages/api/docs/src/paths/share_link.yaml
+++ b/packages/api/docs/src/paths/share_link.yaml
@@ -78,13 +78,14 @@ share-link:
       - name: shareLinkIds
         in: query
         required: false
-        description: An array of share links' ids. Required if shareTokens is not set.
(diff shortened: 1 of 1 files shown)


packages/api/docs/src/paths/share_link.yaml
```diff
@@ -78,13 +78,14 @@ share-link:
...
         required: false
-        description: An array of share links' ids. Required if shareTokens is not set.
+        description: An array of share links' ids. Required if shareTokens is not set. Requires bearer token authentication.
         schema:
           type: array
           items:
             type: string
     summary: Retrieve a share link by ID or token
     security:
+      - {}
```
cecilia-donnelly: For my edification: what is the `security` subitem and what does a bracket here represent?

PR #715, opened 2026-04-13
Fix NewRelic APM

liam-lloyd:
The stela API server no longer appears in NewRelic. This commit updates how we install the New Relic Node.js agent to fix that.


cecilia-donnelly (approved):
This seems reasonable!

PR #726, opened 2026-04-27
Add new location metadata

### LINKED ISSUE #725: Support new location metadata fields
We're working on some updates to the backend related to metadata: https://github.com/PermanentOrg/back-end/issues/862

We need to mirror those changes in Stela as well.  The initial changes should be non-breaking (e.g. add new fields but don't remove the old yet); this gives us time to update clients.

slifty:
This PR updates stela to use the new location metadata types that are being added to the schema / back-end.

Resolves #725


slifty:
I did decide to be non-nonbreaking (but didn't update the issue) and instead just make sure all three PRs are merged "together".

Which is to say, all of copilot's concerns seem to be about that so I hit resolve.  (That said, if human reviewers would prefer to fall back to non-breaking then I'll do so!)

database/base.sql
```diff
@@ -1315,7 +1315,14 @@ CREATE TABLE public.locn (
     status text NOT NULL,
     type text NOT NULL,
     createddt timestamp with time zone,
-    updateddt timestamp with time zone
+    updateddt timestamp with time zone,
```
liam-lloyd: This change should be unnecessary; when we build local databases for stela we run all the migrations on them.

packages/api/docs/src/models/location.yaml
```diff
@@ -5,23 +5,27 @@ location:
...
-      type: string
-    countryCode:
-      type: string
-    displayName:
+    altitudeMeters:
+      type: number
+    precision:
       type: string
+      enum: [approximate, uncertain, unknown]
+    raw:
```
liam-lloyd: I don't think this field should be exposed by the API; if callers needed something that we're keeping in the raw json column I think that would be an indication that we should promote that field to a regular column.

slifty: Agreed!  I had opted to keep it thinking maybe we'd want to be able to write to it, but it makes sense to NOT support that and treat it as an internal column that we may use for future data migrations.

liam-lloyd (approved):

PR #730, opened 2026-04-28
PER-10551: Show specific EDTF errors

cecilia-donnelly:
In order to show users more useful errors, surface those errors from our EDTF package's parser instead of the basic validator.


liam-lloyd (approved):

cecilia-donnelly:
Possibly it would be better to use `parseLevel1` specifically instead of generic `parse`: https://github.com/BobPritchett/edtf-ts/blob/main/docs/guide/parsing.md.

PR #732, opened 2026-04-29
Add new location metadata

### LINKED ISSUE #725: Support new location metadata fields

slifty:
This PR adds our new location fields but this time without removing the deprecated fields; this gives clients time to migrate to using the new values.

Resolves #725


packages/api/src/common/models.ts
```diff
@@ -9,14 +9,26 @@ export enum AccessRole {
...
+	sublocation?: string;
+	city?: string;
+	state?: string;
+	postalCode?: string;
+	country?: string;
+	latitude?: number;
+	longitude?: number;
+	altitudeMeters?: number;
+	precision?: "approximate" | "uncertain" | "unknown";
+	/** @deprecated Use `sublocation` instead. */
```
liam-lloyd: Based on the way the linter interprets deprecated fields, I think it's maybe incorrect to use this tag for interfaces internal to this repo. The point of the tag is to tell users that they shouldn't use this field, but in our case the only user is this repo, which _should_ use this field because we're intentionally still supporting it..

slifty: Ahhh yes that makes sense; fixing now.

liam-lloyd (approved):

PR #733, opened 2026-04-30
PER-10577: Add name to pendingShare object

cecilia-donnelly:
We added this `pendingShare` attribute to records and folders in response to new designs/screens on the mobile apps. When I made the original ticket, I missed that those designs included the name of the invitee. Add that name to both folder and record responses. This is a followup to #688.

### DIFF
--- a/packages/api/src/record/queries/get_records.sql
+++ b/packages/api/src/record/queries/get_records.sql
@@ -111,6 +111,7 @@ aggregated_pending_shares AS (
     ARRAY_AGG(JSONB_BUILD_OBJECT(
       'id', invite_share.invite_shareid::TEXT,
       'email', invite.email,
+      'name', invite.fullname,
       'accessRole', invite_share.accessrole
     )) AS pending_shares_as_json
   FROM invite_share
--- a/packages/api/docs/src/models/share.yaml
+++ b/packages/api/docs/src/models/share.yaml
@@ -20,6 +20,9 @@ pendingShare:
       type: string
     email:
       type: string
+    name:
+      type: string
+      nullable: true
     accessRole:
       type: string
 shareArchive:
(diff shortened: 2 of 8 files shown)


cecilia-donnelly:
@luciancerbu-vsp, adding you as a reviewer here so you can see progress.

packages/api/src/record/queries/get_records.sql
```diff
@@ -111,6 +111,7 @@ aggregated_pending_shares AS (
     ARRAY_AGG(JSONB_BUILD_OBJECT(
       'id', invite_share.invite_shareid::TEXT,
       'email', invite.email,
+      'name', invite.fullname,
```
slifty, line 114: One flag here -- invite.fullname is nullable, but we're saying that `name` will be a string.

I think we already have this problem for `accessRole` and I also suspect that practically speaking fullname won't be null.  That said, this is not fully type safe (and it's at the SQL=> ts layer so it's on us to make sure our queries are doing what we expect).

Three paths forward:

1. It's fine, we will hopefully not be burned by it!
2. We COALESCE with `''` at the SQL level so that if it's null it resolves to an empty string <-- I do think null is more explicit than empty string, but also 🤷🏻 
3. We update the type to be string | null which means clients know they need to potentially handle the null case.

slifty (approved):
Slight type mismatch between what SQL is returning and what we are promising; I'm approving since this is already a category of bug and probably the ideal solution here would be to make fullname a non-nullable field anyway.

That said, technically I believe `name` should be `string | null` in the current state.

luciancerbu-vsp (approved):

packages/api/src/record/queries/get_records.sql
cecilia-donnelly: Thanks! I went with the third option and in the yaml as well - might as well be explicit. :)

PR #744, opened 2026-05-05
Per 10589 identify access copy filetypes better

liam-lloyd:


slifty:
Copilot is thinking really really hard about its review 🤔

liam-lloyd:
The easiest way to test this is to upload a PDF and verify that the access copy generated by Archivematica is correctly identified as a PDF. This works because since PDFs are already in an access format, the access copy will just be a plain copy of the original, and Archivematica stores the original without a file extension.

slifty (approved):

PR #747, opened 2026-05-07
Per 10542 add record copy endpoint

liam-lloyd:


liam-lloyd:
Testing notes:

The permissions requirements for this endpoint are:

- The caller needs Curator permissions on the archive where the record lives

- But if the copy would be created in a public workspace, they actually need Manager permissions

- The destination folder must be in the same archive as the record

- Unless the caller owns the record archive and has at least curator permissions on the destination archive (or manager permissions on the destination if copying to a public place)



This endpoint is expected to create `record`, `folder_link`, `tag_link`, `file`, and `record_file` entries for the copy, but isn't expected to copy the file in S3, populate the new `file` row's URLs, or charge storage. These operations will be handled asynchronously and triggered by the `file.copy` event this endpoint writes to the database.

packages/api/src/record/service.ts
```diff
@@ -94,6 +101,137 @@ export const patchRecord = async (
...
+		if (availableStorage.rows[0] === undefined) {
+			logger.error("Empty response from account_space query");
+			throw new createError.InternalServerError(
+				"Failed to look up account space",
+			);
+		} else if (+availableStorage.rows[0].spaceLeft < recordSize) {
+			throw new createError.BadRequest("Not enough storage to make a copy");
+		}
+
+		const copiedRecordId = await transactionDb
```
slifty: At first glance I had originally interpreted this as "the ID of the original record that was copied" rather than "the ID of the resulting copy of a record". 

Maybe instead of `copiedRecord` which could potentially reference either side, we could use `recordCopy` -- so this would be `recordCopyId`

.sqlfluff
```diff
@@ -2,6 +2,7 @@
 dialect = postgres
 ignore = templating
 templater = placeholder
+exclude_rules = ST03
```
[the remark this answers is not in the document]
liam-lloyd: It is our policy to globally disable rules that don't apply globally, in order to promote a culture of _always_ following the rules that are enabled.

packages/api/src/record/service.ts
```diff
@@ -94,6 +101,137 @@ export const patchRecord = async (
...
+					PrettyFolderType.Public,
+				].includes(destinationFolder.type),
+				callerIp: requestBody.ip,
+				callerUserAgent: userAgent,
+			})
+			.catch((err: unknown) => {
+				logger.error(err);
+				throw new createError.InternalServerError("Failed to copy record");
+			});
+		if (copiedRecordId.rows[0] === undefined) {
```
slifty: This is a little awkward as written since `copiedRecordId` is named as though it would be a primitive, but it's actually a sql result.

What if we instead did:

```
const [{ recordId: recordCopyId }] = await transactionDb...
```

And then this could be`if (recordCopyId === undefined) { ... })

And later you don't have to do the whole .rows[0] dance when returning

packages/api/src/record/service.ts
```diff
@@ -4,16 +4,23 @@ import { db } from "../database";
...
 import { AccessRole } from "../access/models";
 import { shareLinkService } from "../share_link/service";
 import type { ShareLink } from "../share_link/models";
+import { getFolders } from "../folder/service";
+import { type Folder, PrettyFolderType } from "../folder/models";
 
 export const getRecords = async (requestQuery: {
 	recordIds: string[] | undefined;
-	archiveId: string | undefined;
+	archiveId?: string | undefined;
```
slifty, line 23: This is being marked as optional now but I'm not sure why!

packages/api/src/record/service.ts
```diff
@@ -94,6 +101,137 @@ export const patchRecord = async (
...
+const validateHasPermissionToCopyRecord = async (
+	record: ArchiveRecord,
+	destinationFolder: Folder,
+	authenticatedEmail: string,
+): Promise<void> => {
+	const recordArchiveAccessRole = await getArchiveAccessRole(
+		record.archive.id,
+		authenticatedEmail,
+	);
+	const folderArchiveAccessRole = await getArchiveAccessRole(
```
slifty: Could we name these two variables as `originArchiveAccessRole` and `destinationArchiveAccessRole`?

I think that's the "meaning" here / that prevents the need to mentally remember that the "record" is associated with the origin archive and the "folder" is associated with the destination.

It might even be helpful to, at the top of this function, just define the archive Ids directly e.g. 

```
const { archive: originArchive } = record;
const { archive: destinationArchive } = destinationFolder;
```

And then later checks are slightly more self-documenting:

```
if ( originArchive.id !== destinationArchive.id )
```

slifty (changes requested):
Some naming suggestions -- the only one I'd truly ask for is the `copiedRecordId` since english is annoying like that.

packages/api/src/record/service.ts
liam-lloyd: Typescript won't let me do this destructuring when it's still possible the array might be empty, but I've renamed `copiedRecordId` to `recordCopyResult`

packages/api/src/record/service.ts
liam-lloyd: `createRecordCopy` calls `getRecords` and I wanted to be able to do that without passing an `archiveId`. It seemed annoying to have to pass `undefined` explicitly, though I could be convinced that it's a good idea to require that.

packages/api/src/record/service.ts
liam-lloyd: Actually, looks like I can do `const { rows: [ recordCopy ] } = ...`

packages/api/src/record/service.ts
slifty: Ah that sounds fine with me!  May want to remove `| undefined` since I think that is now redundant, but not a big deal :D

slifty (approved):
Looks great!

PR #754, opened 2026-05-11
Allow location editing for `record` and `folder` entities

### LINKED ISSUE #743: Add `POST` endpoint for locations
This issue is for the addition of a `POST` / `PUT` endpoint to create locations which are used by records and folders.

Related to https://permanent.atlassian.net/issues?jql=textfields%20~%20%22stela%20location*%22&selectedIssue=PER-10421

slifty:
This PR updates the stela folder / record endpoints to support direct editing of the location associated with a given entity.

When editing locations via these endpoints, clients will be interacting with locations as though they are 1:1 to a given folder or record.  In implementation stela is (for now) mapping to the current data structure.

Once clients are using this endpoint we will be able to move forward on the ERD migration described in #746 

Resolves #743


packages/api/src/record/service.ts
```diff
@@ -66,32 +68,58 @@ const validateCanPatchRecord = async (
...
+): Promise<string> => {
+	await validateCanPatchRecord(recordId, recordData.emailFromAuthToken);
+
+	return await db.transaction(async (transactionDb) => {
+		let { locationId } = recordData;
+		if (recordData.location !== undefined) {
+			const currentLocnId =
+				(await getRecordLocnId(recordId, transactionDb)) ?? null;
+			locationId = BigInt(
+				await upsertLocation(currentLocnId, recordData.location, transactionDb),
```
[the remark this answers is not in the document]
slifty: I think this is just the nature of the "migration" path.

packages/api/src/folder/service.ts
```diff
@@ -210,31 +212,60 @@ const validateCanPatchFolder = async (
...
+		let locnid: string | null = null;
+		if (folderData.location !== undefined) {
+			const currentLocnId =
+				(await getFolderLocnId(folderId, transactionDb)) ?? null;
+			locnid = await upsertLocation(
+				currentLocnId,
+				folderData.location,
+				transactionDb,
+			);
+		}
```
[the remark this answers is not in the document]
slifty: Again -- for now I think this is just how locations will work.  We expect to migrate to 1:1 shortly.  I would recommend we ignore this risk / I don't think it's realistically going to manifest as a problem.

packages/api/src/record/queries/get_record_locnid.sql
```diff
@@ -0,0 +1,3 @@
+SELECT locnid::text AS locnid
+FROM record
+WHERE recordid = :recordId;
```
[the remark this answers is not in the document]
slifty: This is an interesting suggestion -- I will want to look into whether adding `FOR UPDATE` causes any unexpected behavior

packages/api/docs/src/models/location.yaml
```diff
@@ -44,3 +44,55 @@ location:
...
+      deprecated: true
+    county:
+      type: string
+      deprecated: true
+    countryCode:
+      type: string
+      deprecated: true
+    displayName:
+      type: string
+      deprecated: true
```
liam-lloyd: Do we need to include deprecated fields in the update requests?

packages/api/src/location/service.ts
```diff
@@ -0,0 +1,84 @@
+import createError from "http-errors";
```
liam-lloyd, line 1: I think these changes reveal that the `common` directory was a mistake. I think all this location code should live in a `location` directory and the `AccessRole` enum should move to the `access` directory

packages/api/src/common/service.ts
```diff
@@ -0,0 +1,84 @@
...
+			logger.error(err);
+			throw new createError.InternalServerError("Failed to update location");
+		});
+	if (result.rows[0] === undefined) {
+		throw new createError.NotFound(`Location ${locnid} not found`);
+	}
+};
+
+export const upsertLocation = async (
+	currentLocnId: string | null,
```
liam-lloyd: I think we should spell out "Location" in this variable name

packages/api/src/common/service.ts
```diff
@@ -0,0 +1,84 @@
...
+	const [inserted] = rows;
+	if (inserted === undefined) {
+		throw new createError.InternalServerError("Failed to create location");
+	}
+	const { locnid } = inserted;
+	return locnid;
+};
+
+export const updateLocation = async (
+	locnid: string,
```
liam-lloyd: I think we should spell out "location" in this variable name

packages/api/src/common/service.ts
```diff
@@ -0,0 +1,84 @@
...
+	currentLocnId: string | null,
+	location: LocationInput,
+	client: TinyPg = db,
+): Promise<string> => {
+	if (currentLocnId === null) {
+		return await insertLocation(location, client);
+	}
+	await updateLocation(currentLocnId, location, client);
+	return currentLocnId;
+};
```
liam-lloyd: There's something slightly un-upsert-like about this function, in that if I gave it a location ID that didn't exist it would fail rather than creating a new location with that ID. That would be a bad behavior, so it's good that it won't do it, but it makes me wonder whether the caller shouldn't just figure out whether they have a location ID and call insert or update accordingly. Or maybe I'm just being overly pedantic about what upsert means?

packages/api/src/folder/service.ts
```diff
@@ -210,31 +212,60 @@ const validateCanPatchFolder = async (
...
-			setDisplayEndDateToNull: folderData.displayEndDate === null,
-			displayTime: folderData.displayTime,
-			setDisplayTimeToNull: folderData.displayTime === null,
 		})
 		.catch((err: unknown) => {
 			logger.error(err);
-			throw new createError.InternalServerError("Failed to update folder");
+			throw new createError.InternalServerError("Failed to look up folder");
 		});
+	return result.rows[0]?.locnid;
```
liam-lloyd: If rows[0] is undefined, that should be a 400 error shouldn't it?

packages/api/src/record/service.ts
```diff
@@ -66,32 +68,58 @@ const validateCanPatchRecord = async (
...
-			setDescriptionToNull: recordData.description === null,
-			displayTime: recordData.displayTime,
-			setDisplayTimeToNull: recordData.displayTime === null,
 		})
 		.catch((err: unknown) => {
 			logger.error(err);
-			throw new createError.InternalServerError("Failed to update record");
+			throw new createError.InternalServerError("Failed to look up record");
 		});
+	return result.rows[0]?.locnid;
```
liam-lloyd: Like in folder, it seems like rows[0] === undefined should be a 400 error

packages/api/src/record/validators.test.ts
```diff
@@ -60,4 +60,127 @@ describe("validatePatchRecordRequest", () => {
...
+					countryCode: "FR",
+					displayName: "Jean Valjean's House",
+				},
+			});
+		} catch (err) {
+			error = err;
+		} finally {
+			expect(error).toBeNull();
+		}
+	});
```
liam-lloyd, line 185: These are redundant with the location tests in the folder validator; maybe we should test the location validation in the shared `location` directory?

liam-lloyd (commented):
I didn't comment on it in all the places it exists, but I don't think we should ever use the "locn" abbreviation except in SQL files where we have to because that is the existing column name

packages/api/docs/src/models/location.yaml
slifty: It depends on if we're OK with a window of mismatch between clients.

For instance, if we update the web app to write to this and NOT populate the old values, then locations in the mobile app will appear incomplete until such a time that the mobile app has migrated to read the new values.

packages/api/src/common/service.ts
slifty: Whoops these slipped by me -- agreed in all cases

packages/api/src/record/validators.test.ts
slifty: I wouldn't advocate for that -- they appear redundant but the point here is that the concept of a `location` as an entity is going away.  These are, really, testing the location attributes of a folder (and likewise we need to test the location attributes of a record).  Similar to how both folder and record both have a display name but we would redundantly test both.

packages/api/src/common/service.ts
slifty: I will take a look and update accordingly

That said I want to highlight that the entire concept of "upsert for location" is a dead-on-arrival shim / we're going to be blowing most of this away in 2 weeks when we migrate the locations table.

This is not a traditional upsert though, by any means.   That said, there isn't a proper term for what we're doing since what we're doing is temporarily simulating a 1:1 relationship a cross a 1:n schema design and nobody would *choose* that kind of thing 😂 

I'm not saying we should add incorrect code though!

packages/api/src/record/validators.test.ts
liam-lloyd: I was thinking of it as more analogous to middleware, where we test on each endpoint that the correct middleware gets called but the more exhaustive tests of the middleware live with the middleware. But due to the nature of how Joi schemas are built maybe it's not really possible to do that here, and given that then I agree it's correct to test in both places.

packages/api/docs/src/models/location.yaml
cecilia-donnelly: I talked to @slifty about this and determined that this is the right approach. It will likely be around six months before the mobile apps do their location update, which is a natural time for them to switch to using the new fields. That said, I will make a note about ensuring that we *actually* deprecate and remove the old fields at that point.

slifty: One thing I will explore is whether it would be easy to have stela translate the new fields to the deprecated values on write, which would mean stela doesn't have to expose the old ones.

slifty: All right -- I am using that approach of "values are passed only using new fields, but written to the new AND legacy fields".  This keeps the API itself cleaner / once migrated to it, clients won't have to do anything (and once all clients are migrated we can just remove all the gnarly shims that attempt to translate the new fields to the old fields)

packages/api/src/record/validators.test.ts
slifty: Closing the loop here towards short term resolution: I do think that once the dust settles on the location migration this is worth revisiting.

Once we're in the final state we might find that we either (A) want to think of the location of a folder is a distinct concept from a location of a record (similar to how like display name exists on both but they are tested separately), or (B) think of location as a complex object that has common tests as you're describing (and heck maybe we will want to store it as a jsonb in sql / a single attribute).

For now I'll leave it as is since it sounds somewhat acceptable and the work in this area is in progress!

packages/api/src/location/service.ts
```diff
@@ -0,0 +1,136 @@
...
+	}
+	return inserted.locationId;
+};
+
+export const updateLocation = async (
+	locationId: string,
+	location: LocationInput,
+	client: TinyPg = db,
+): Promise<void> => {
+	// `displayName` is a concatenation of multiple modern fields, so a partial
```
slifty: Apparently display name is not actually populated by the php code base... so we don't need to start populating it now.  Given this, we can simplify this function significantly.

packages/api/src/folder/service.ts
```diff
@@ -210,31 +212,75 @@ const validateCanPatchFolder = async (
...
+			},
+		)
+		.catch((err: unknown) => {
+			logger.error(err);
+			throw new createError.InternalServerError("Failed to look up folder");
+		});
+	const { rows } = result;
+	const [row] = rows;
+	if (row === undefined) {
+		throw new createError.BadRequest(`Folder ${folderId} not found`);
```
liam-lloyd: Is this bad request or not found?

packages/api/src/record/service.ts
```diff
@@ -66,32 +68,77 @@ const validateCanPatchRecord = async (
...
+			},
+		)
+		.catch((err: unknown) => {
+			logger.error(err);
+			throw new createError.InternalServerError("Failed to look up record");
+		});
+	const { rows } = result;
+	const [row] = rows;
+	if (row === undefined) {
+		throw new createError.BadRequest(`Record ${recordId} not found`);
```
liam-lloyd: Same question about bad request vs. not found

liam-lloyd (changes requested):
It looks to me like the codecov complaints here are legitimate; we're missing a lot of DB query failure cases around dealing with locations.

packages/api/src/folder/service.ts
slifty: Good catch ty!

slifty:
@liam-lloyd codecov is saying the *project* percentage went down even though the patch coverage is 100%.  I don't quite understand how that works...  It suggests to me that something is weird about how project (or patch) coverage % is being calculated.

Codecov aside though I'm happy to add more tests!

slifty:
I added some direct tests to the location service to cover some of the edge db conditions (which I think maybe also indirectly addresses some of your earlier thoughts about when location tests belong in the folder / record vs in a location-specific place?)

liam-lloyd:
> @liam-lloyd codecov is saying the _project_ percentage went down even though the patch coverage is 100%. I don't quite understand how that works... It suggests to me that something is weird about how project (or patch) coverage % is being calculated.
> 
> Codecov aside though I'm happy to add more tests!

Definitely something is weird about how the percentages are calculated, but I think it is flagging real holes in the test coverage of this PR, because they're places like this: 
image
image

It is quite odd that the patch level check doesn't see these, and also that the project level check sees them as "indirect".

slifty:
@liam-lloyd argh ok sorry -- I was too trusting in the patch report (usually I'd see those missing lines in a patch report)

I'll separately open an issue to investigate coverage reporting so we get more accurate info in future, and then will ping you when this is all sorted :P

liam-lloyd (approved):

PR #757, opened 2026-05-14
Per 10543 add file copy lambda

liam-lloyd:
This is best tested in conjunction with the copy endpoint added in https://github.com/PermanentOrg/stela/pull/747; ideally by the time you are testing this that has been merged. Then you can test by simply hitting that endpoint and checking that after a short time to allow async processes to run:
- The copy has an original file that you can download
- The copy has an access copy that you can download
- The copy has a thumbnail
Right now it looks to me like the web-app isn't displaying thumbnails when only an Archivematica thumbnail exists, which is the expected end state for these copies.

### DIFF
--- a/packages/file_copier/src/index.ts
+++ b/packages/file_copier/src/index.ts
@@ -1,0 +1,137 @@
+import type { SQSHandler, SQSEvent, SQSRecord } from "aws-lambda";
+import { S3Client, CopyObjectCommand } from "@aws-sdk/client-s3";
+import * as Sentry from "@sentry/aws-serverless";
+import { constructSignedCdnUrl, validateSqsMessage } from "@stela/s3-utils";
+import { logger } from "@stela/logger";
+import { validateFileCopyEvent } from "./validator";
+import { db } from "./database";
+
+export const extractFileDataFromFileCopyMessage = (
+	message: SQSRecord,
+): {
+	eventId: string;
+	originalFileId: string;
+	newFileId: string;
+	originalFileCloudPath: string;
+} => {
+	const { body } = message;
+	const parsedBody: unknown = JSON.parse(body);
+	if (!validateSqsMessage(parsedBody)) {
+		logger.error(`Invalid message body: ${body}`);
+		throw new Error("Invalid message body");
+	}
+	const parsedMessage: unknown = JSON.parse(parsedBody.Message);
+	validateFileCopyEvent(parsedMessage);
+
+	return {
+		eventId: parsedMessage.id,
+		originalFileId: parsedMessage.body.file.fileid.toString(),
+		newFileId: parsedMessage.body.newFile.fileid.toString(),
+		originalFileCloudPath: parsedMessage.body.file.cloudpath,
+	};
+};
+
+const generateNewFileCloudPath = (
+	originalFileCloudPath: string,
+	originalFileId: string,
+	newFileId: string,
+): string => {
+	const originalFileCloudPathWithNewFileIds = originalFileCloudPath.replace(
+		new RegExp(`(?<=^|/)${originalFileId}(?=/|$)`, "vg"),
+		newFileId,
+	);
+
+	// Currently, cloudPaths for original files are in the format "originals/fileId/fileId".
+	// This allows us to generate a cloudPath for the copy by simply replacing the original
+	// file's ID with the copy's ID everywhere it appears in the cloudPath. However, some
+	// older files have cloudPaths of simply "fileId". In those cases, we to generate a cloudPath
+	// for the copy in the new format. Since in the previous step we replaced all instances of
+	// the original fileId with the copy's fileId, this is usually just a matter of adding
+	// "originals/newFileId/" as a prefix. However, in developers' local environments cloudPaths
+	// in the old and new formats have a prefix of "_NameOfDeveloper/", so if we're converting an
+	// old-format cloudPath to a new-format cloudPath we check such a prefix and include it in the
+	// generated cloudPath.
+	return originalFileCloudPathWithNewFileIds.includes("originals")
+		? originalFileCloudPathWithNewFileIds
+		: originalFileCloudPathWithNewFileIds.replace(
+				/^(?<localPrefix>.*\/)?/v,
+				`$1originals/${newFileId}/`,
+			);
+};
+
+const getFileUploadName = async (fileId: string): Promise<string> => {
+	const newFile = await db
+		.sql<{ uploadFileName: string }>("queries.get_file_upload_name", {
+			fileId,
+		})
+		.catch((err: unknown) => {
+			logger.error(err);
+			throw err;
+		});
+	if (newFile.rows[0] === undefined) {
+		const errorMessage = "Failed to look up new file";
+		logger.error(errorMessage);
+		throw new Error(errorMessage);
+	}
+
+	return newFile.rows[0].uploadFileName;
+};
+
+export const handler: SQSHandler = Sentry.wrapHandler(
+	async (event: SQSEvent) => {
+		const s3Client = new S3Client({
+			region: process.env["AWS_REGION"] ?? "",
+		});
+		await Promise.all(
+			event.Records.map(async (message) => {
+				const { eventId, originalFileId, newFileId, originalFileCloudPath } =
+					extractFileDataFromFileCopyMessage(message);
+
+				const newFileCloudPath = generateNewFileCloudPath(
+					originalFileCloudPath,
+					originalFileId,
+					newFileId,
+				);
+
+				await s3Client
+					.send(
+						new CopyObjectCommand({
+							Bucket: process.env["S3_BUCKET"] ?? "",
+							CopySource: encodeURI(
+								`${process.env["S3_BUCKET"] ?? ""}/${originalFileCloudPath}`,
+							),
+							Key: newFileCloudPath,
+						}),
+					)
+					.catch((err: unknown) => {
+						logger.error(err);
+						throw err;
+					});
+
+				const cdnExpirationTime = new Date();
+				cdnExpirationTime.setFullYear(cdnExpirationTime.getFullYear() + 1);
+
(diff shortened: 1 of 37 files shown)


liam-lloyd:
If you want to test this locally, you'll need https://github.com/PermanentOrg/devenv/pull/90

terraform/test_cluster/file_copier_dev_lambda.tf
```diff
@@ -0,0 +1,126 @@
...
+          "sqs:ReceiveMessage",
+          "sqs:DeleteMessage",
+          "sqs:GetQueueAttributes",
+          "s3:GetObject",
+          "s3:PutObject",
+        ]
+        Effect   = "Allow"
+        Resource = ["*", aws_sqs_queue.file_copier_dev_queue.arn]
+      },
+    ]
```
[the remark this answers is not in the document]
liam-lloyd: This is how we set permissions for all our lambdas

packages/file_copier/src/index.ts
```diff
@@ -0,0 +1,106 @@
...
+		const s3Client = new S3Client({
+			region: process.env["AWS_REGION"] ?? "",
+		});
+		await Promise.all(
+			event.Records.map(async (message) => {
+				const { eventId, originalFileId, newFileId, originalFileCloudPath } =
+					extractFileDataFromFileCopyMessage(message);
+
+				const originalFileCloudPathWithNewFileIds =
+					originalFileCloudPath.replaceAll(originalFileId, newFileId);
```
slifty: I think this isn't specific enough, since the id string could appear in the file path in other ways -- e.g. `originals/100/file_v100_thumb.jpg, original id 100 → originals/101/file_v101_thumb.jpg`

packages/file_copier/src/queries/update_file_and_send_event.sql
```diff
@@ -0,0 +1,218 @@
...
+    cloudpath = :cloudPath,
+    fileurl = :fileUrl,
+    downloadurl = :downloadUrl,
+    urldt = :urlExpirationTimestamp
+  WHERE
+    fileid = :newFileId
+  RETURNING fileid
+)
+
+INSERT INTO event (
```
slifty, line 15: This particular insert isn't idempotent (and SQS commits to "[at least once](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/standard-queues-at-least-once-delivery.html)" which means in theory this could be triggered more than once).

Can we do a duplicate key guard somehow?

packages/file_copier/src/index.ts
```diff
@@ -0,0 +1,106 @@
...
+					: originalFileCloudPathWithNewFileIds.replace(
+							/^(?<localPrefix>.*\/)?/v,
+							`$1originals/${newFileId}/`,
+						);
+
+				await s3Client
+					.send(
+						new CopyObjectCommand({
+							Bucket: process.env["S3_BUCKET"] ?? "",
+							CopySource: `${process.env["S3_BUCKET"] ?? ""}/${originalFileCloudPath}`,
```
slifty: I don't remember for certain if this is needed here, but would it be appropriate to `encodeURIComponent` for the originalFileCloudPath?

packages/file_copier/src/index.ts
```diff
@@ -0,0 +1,106 @@
...
+					? originalFileCloudPathWithNewFileIds
+					: originalFileCloudPathWithNewFileIds.replace(
+							/^(?<localPrefix>.*\/)?/v,
+							`$1originals/${newFileId}/`,
+						);
+
+				await s3Client
+					.send(
+						new CopyObjectCommand({
+							Bucket: process.env["S3_BUCKET"] ?? "",
```
slifty, line 99: Do we want this to silently allow a lack of S3_BUCKET configuration?

packages/file_copier/src/index.ts
```diff
@@ -0,0 +1,106 @@
...
+		originalFileCloudPath: parsedMessage.body.file.cloudpath,
+	};
+};
+
+export const handler: SQSHandler = Sentry.wrapHandler(
+	async (event: SQSEvent) => {
+		const s3Client = new S3Client({
+			region: process.env["AWS_REGION"] ?? "",
+		});
+		await Promise.all(
```
slifty, line 85: Something claude flagged related to batch failures which I think will probably end up being out of scope for this PR.  In the context of batches a single message failure rejects the entire batch and SQS would retry all of them.

Claude suggested that we could do better by adding support for batch failure reporting:

```
  export const handler: SQSHandler = Sentry.wrapHandler(
    async (event: SQSEvent) => {
      const s3Client = new S3Client({ region: process.env["AWS_REGION"] ?? "" });
      const results = await Promise.allSettled(
        event.Records.map(async (message) => {
          // ... existing per-message logic, but throw on failure ...
          // (return message.messageId from .catch handlers, or just let throw propagate)
        }),
      );
      const batchItemFailures = results
        .map((r, i) =>
          r.status === "rejected"
            ? { itemIdentifier: event.Records[i]!.messageId }
            : null,
        )
        .filter((x): x is { itemIdentifier: string } => x !== null);
      return { batchItemFailures };
    },
  );
```

Which apparently would also require some terraform changes to handle batch item failure reports:

```
  resource "aws_lambda_event_source_mapping" "file_copier_dev_event_source_mapping" {
    event_source_arn                   = aws_sqs_queue.file_copier_dev_queue.arn
    function_name                      = aws_lambda_function.file_copier_dev_lambda.arn
    batch_size                         = 10
    maximum_batching_window_in_seconds = 0
    function_response_types            = ["ReportBatchItemFailures"]
  }
```

---

If this was a pattern we wanted to actually pursue I think it would apply to other lambdas as well.

Anyway, thought I'd share in case it was worth a followup issue!  Not suggesting a change in this PR.

packages/file_copier/src/index.ts
```diff
@@ -0,0 +1,106 @@
...
+			event.Records.map(async (message) => {
+				const { eventId, originalFileId, newFileId, originalFileCloudPath } =
+					extractFileDataFromFileCopyMessage(message);
+
+				const originalFileCloudPathWithNewFileIds =
+					originalFileCloudPath.replaceAll(originalFileId, newFileId);
+				const newFileCloudPath = originalFileCloudPathWithNewFileIds.includes(
```
slifty: This kind of check I think is helpful to put behind a named function for legibility to help signal what is actually being checked from a logical meaning perspective (e.g. `isCloudPathForOriginalFile` or something)

packages/file_copier/src/index.ts
```diff
@@ -0,0 +1,106 @@
...
+					"originals",
+				)
+					? originalFileCloudPathWithNewFileIds
+					: originalFileCloudPathWithNewFileIds.replace(
```
slifty: Similarly, and I realize this is maybe code style preference so ignore me if you disagree, I think it would be helpful for this transformation to happen in a named function so reading the code I can tell what is *intended* (e.g. is this `convertOriginalFileLocalPathToCloudPath()`?)

packages/file_copier/src/index.ts
```diff
@@ -0,0 +1,106 @@
...
+							CopySource: `${process.env["S3_BUCKET"] ?? ""}/${originalFileCloudPath}`,
+							Key: newFileCloudPath,
+						}),
+					)
+					.catch((err: unknown) => {
+						logger.error(err);
+						throw err;
+					});
+
+				const newFile = await db
```
slifty: Could we put this (and the check) into a function that promises to return the new file (or throw, as it does already)?

That would help shorten up the handler a bit / consolidate that subset of the work.  (Also then we can unit test if it we wanted to)

slifty (changes requested):
The big thing to fix I think is that potential "we re-write more than just the id in the path" item + the idempotency

packages/file_copier/src/index.ts
liam-lloyd: This does work in our currently system because the `cloudPath`s of original files are very constrained; they are either `fileId` or `originals/fileId/fileId`. But I can harden this by looking only for fileIds sandwiched by some combination of ^, $, and \/ (<- interpret as regex)

packages/file_copier/src/index.ts
liam-lloyd: This line does what I want; but it looks like I do need to add some more env var requirements to `env.ts` (which will cause missing env vars to generate errors on startup; the second half of the coalesce here should never occur but is required to make Typescript happy)

packages/file_copier/src/queries/update_file_and_send_event.sql
liam-lloyd: I can make sure the event only gets written if the update actually writes new values

packages/file_copier/src/index.ts
liam-lloyd: Under current conditions, it shouldn't be possible for `originalFileCloudPath` to include characters that we'd need to escape

packages/file_copier/src/index.ts
liam-lloyd: This seems like a good idea, perhaps for the next pothole sprint.

packages/file_copier/src/index.ts
liam-lloyd: While I'm normally in favor of verbose names, `wasOriginalFileUploadedAfterWeStartedUsingTheOriginalsDirectoryInS3` strikes me as too extreme, so I'll add a comment

packages/file_copier/src/index.ts
slifty: We talked through this -- I'm fine as is *but* defensively it's still worth encoding in case that external condition changes (since that constraint isn't determined or maintained within this component, so if it ever changed we likely wouldn't think to come back here)

packages/file_copier/src/index.ts
slifty: I'll spin open a github issue for it just so we track it *somewhere*

slifty (approved):
Thank you for the iteration!  Looking good.

PR #761, opened 2026-05-19
Add GET /accounts endpoint

liam-lloyd:
We're working on an admin tool that will allow administrators to add or remove storage from an account. That tool requires the ability to look up accounts by email. To enable that, this commit adds a GET /accounts endpoint. It is presently restricted to admins only.

### DIFF
--- a/packages/api/docs/src/paths/account.yaml
+++ b/packages/api/docs/src/paths/account.yaml
@@ -41,6 +41,57 @@ account:
         $ref: "../errors.yaml#/400"
       "500":
         $ref: "../errors.yaml#/500"
+  get:
+    summary: Get account objects. Requires admin authentication.
+    parameters:
+      - name: accountIds
+        in: query
+        required: false
+        description: |
+          An array of account IDs to retrieve.
+          If this parameter is provided, accountEmails must not be provided
+        schema:
+          type: array
+          items:
+            type: string
+      - name: accountEmails
+        in: query
+        required: false
+        description: |
+          An array of account emails to retrieve.
+          If this parameter is provided, accountIds must not be provided
+        schema:
+          type: array
+          items:
+            type: string
+    operationId: getAccount
+    security:
+      - bearerHttpAuthentication: []
+    tags:
+      - accounts
+      - admin
+    responses:
+      "200":
+        description: >
+          The accounts requested.
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                items:
+                  type: array
+                  items:
+                    $ref: "../models/account.yaml#/account"
+                pagination:
+                  $ref: "../models/pagination_metadata.yaml#/pagination"
+      "400":
+        $ref: "../errors.yaml#/400"
+      "401":
+        $ref: "../errors.yaml#/401"
+      "500":
+        $ref: "../errors.yaml#/500"
+
 storage-adjustment:
   post:
     parameters:
--- a/packages/api/src/account/service.ts
+++ b/packages/api/src/account/service.ts
@@ -12,7 +12,6 @@ import {
 	type UpdateTagsRequest,
 	type SignupDetails,
 	type GetAccountArchiveResult,
-	type GetCurrentAccountArchiveResult,
 	type LeaveArchiveRequest,
 	type CreateStorageAdjustmentRequest,
 	type StorageAdjustment,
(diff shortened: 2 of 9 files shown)


packages/api/docs/src/paths/account.yaml
```diff
@@ -41,6 +41,57 @@ account:
...
         $ref: "../errors.yaml#/500"
+  get:
+    summary: Get account objects. Requires admin authentication.
+    parameters:
+      - name: accountIds
+        in: query
+        required: false
+        description: |
+          An array of account IDs to retrieve.
+          Exactly one of accountIds or accountEmails must be provided
```
cecilia-donnelly: It took me a little time to understand this! I thought it meant that we were expecting only one element in the array. So this means that I can't ask for something like `['someone@example.com', '12345', 'anotheremail@example.com', '2233', '4343']`. Is that right? This is very concise, but for me something longer might be easier to understand. Maybe "The array must either be all email addresses or all account ids"?

cecilia-donnelly: Possibly that should go in the summary instead of the description of the fields? Or maybe for this one, "must not be provided if `accountEmails` is provided" or something like that?

cecilia-donnelly: Thinking out loud, but really what we're trying to avoid is `get/accounts/accountIds=['123', '223', '323']&accountEmails=['account1@example.com', 'account2@example.com']`. Right?

cecilia-donnelly (commented):
I'm excited about this! Just leaving a comment about the docs since I see the tests are still failing so this probably isn't ready for review. Thank you, @liam-lloyd .

packages/api/docs/src/paths/account.yaml
liam-lloyd: Yes, that's right. I've updated the description to something close to your suggestion in the second comment.

liam-lloyd:
The second commit is included here to appease codecov

packages/api/docs/src/paths/account.yaml
```diff
@@ -41,6 +41,57 @@ account:
...
+        schema:
+          type: array
+          items:
+            type: string
+      - name: accountEmails
+        in: query
+        required: false
+        description: |
+          An array of account emails to retrieve.
+          If this parameter is provided, accountIds must not be provided
```
cecilia-donnelly, line 62: Thanks, this is clear!

packages/api/src/account/service.ts
```diff
@@ -16,8 +16,90 @@ import type {
...
+	const accountIds = Array.isArray(query.accountIds)
+		? query.accountIds
+		: query.accountIds === undefined
+			? []
+			: [query.accountIds];
+	const accountEmails = Array.isArray(query.accountEmails)
+		? query.accountEmails.map((e) => e.toLowerCase())
+		: query.accountEmails === undefined
+			? []
+			: [query.accountEmails.toLowerCase()];
```
cecilia-donnelly, line 98: So if they do send both then they only get results from the id array they sent, am I reading that right? I think (but could be persuaded otherwise) that we should throw if they send both. Otherwise the results might be misleading, if the user thinks the results represent the findings from both arrays.

packages/api/src/account/validators.ts
```diff
@@ -89,6 +92,28 @@ export const validateCreateStorageAdjustmentRequest: (
...
+		.keys({
+			...paginationFields,
+			accountIds: Joi.alternatives()
+				.try(Joi.string(), Joi.array().items(Joi.string()))
+				.optional(),
+			accountEmails: Joi.alternatives()
+				.try(Joi.string().email(), Joi.array().items(Joi.string().email()))
+				.optional(),
+		})
+		.xor("accountIds", "accountEmails")
```
cecilia-donnelly, line 110: Ah perfect, the check is already here.

packages/api/src/account/service.ts
cecilia-donnelly: I see it in the validator - thanks!

cecilia-donnelly (commented):
This currently doesn't consider deleted or test accounts (so far as I can tell). I think we should definitely include those fields (`type` and `status`, I believe) in the response, so admins can see whether an account is deleted or a test account. In a later stage, maybe we could optionally exclude those?

I could also see an argument that we should always exclude deleted accounts, at least, since this is specifically intended to be used for storage adjustments. I expect this endpoint might be used for something else later, though, so that doesn't feel necessary.

liam-lloyd:
> This currently doesn't consider deleted or test accounts (so far as I can tell). I think we should definitely include those fields (`type` and `status`, I believe) in the response, so admins can see whether an account is deleted or a test account. In a later stage, maybe we could optionally exclude those?
> 
> I could also see an argument that we should always exclude deleted accounts, at least, since this is specifically intended to be used for storage adjustments. I expect this endpoint might be used for something else later, though, so that doesn't feel necessary.

I think excluding deleted accounts here is correct, I'll make that update (and add type and status to the account object)

cecilia-donnelly (approved):

PR #771, opened 2026-06-06
Allow metadata files with a single MediaInfo track

liam-lloyd:
Currently, our metadata attacher lambda expects a MediaInfo object in the metadata file to contain multiple tracks. This isn't always the case, so this commit updates the lambda to accept metadata files with only one track.


cecilia-donnelly (approved):
Okay, this makes perfect sense! Thanks!

I do worry that the overall extraction is a bit fragile, given that it looks like there are several different specific metadata errors (though they seem to be the same general type-mismatch problem) in Sentry. What's your sense of how resilient the process is at the moment? Do individual types of metadata fail individually, or if one fails does the whole process stop?

cecilia-donnelly:
I didn't run this, since the code change struck me as extremely straightforward and I didn't immediately find an obvious way to create a test file (video/audio, not just the metadata as in the fixture) with just one of these tracks.

liam-lloyd:
> Okay, this makes perfect sense! Thanks!
> 
> I do worry that the overall extraction is a bit fragile, given that it looks like there are several different specific metadata errors (though they seem to be the same general type-mismatch problem) in Sentry. What's your sense of how resilient the process is at the moment? Do individual types of metadata fail individually, or if one fails does the whole process stop?

The intention is for things to fail individually (i.e. for fields to be optional), but in cases like this where the structure surprises us the whole thing will fail.

PR #776, opened 2026-06-12
Add EDTF Level 2 support

liam-lloyd:
We would like our users to be able to use all the features of EDTF, so this commit updates our record and folder endpoints to support EDTF Level 2. It also adds tests to cover handling of Level 2 features by the database trigger that calculates minimum possible timestamps from EDTF statements.

### DIFF
--- a/packages/api/Dockerfile
+++ b/packages/api/Dockerfile
@@ -9,6 +9,7 @@ COPY tsconfig.build.json ./
 COPY tsconfig.json ./
 COPY jest.config.js ./
 COPY packages ./packages
+COPY patches ./patches
 
 RUN npm install
 RUN npm install -ws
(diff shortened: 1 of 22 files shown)


liam-lloyd:
Requires https://github.com/PermanentOrg/back-end/pull/884, which is why the tests are failing

packages/api/Dockerfile
```diff
@@ -38,6 +40,7 @@ COPY --from=builder /usr/local/apps/stela/packages/permanent_models/package.json
 COPY --from=builder /usr/local/apps/stela/package.json ./package.json
 COPY --from=builder /usr/local/apps/stela/package-lock.json ./package-lock.json
 
+RUN npm install
```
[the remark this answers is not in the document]
liam-lloyd: This does in fact work as is, but I'll add the --workspaces=false for clarity

packages/api/src/folder/validators.ts
```diff
@@ -4,6 +4,8 @@ import type { PatchFolderRequest } from "./models";
 import { fieldsFromUserAuthentication } from "../validators";
 import { locationInputSchema } from "../location/validators";
 
+const SUPPORTED_EDTF_LEVEL = 2;
```
slifty: Ironically this name, while avoiding a magic number, makes it harder to know *what* edtf level we're supporting at the point we are actually interacting with EDTF...

As opposed to 

`EDTF_LEVEL_2 = 2` which means later we're being explicit that we are parsing at `parseEDTF(value, EDTF_LEVEL_2`);`

This is a total nit though so only change if you feel inspired.

packages/api/src/record/validators.ts
```diff
@@ -4,6 +4,8 @@ import type { CreateRecordCopyRequest, PatchRecordRequest } from "./models";
 import { fieldsFromUserAuthentication } from "../validators";
 import { locationInputSchema } from "../location/validators";
 
+const SUPPORTED_EDTF_LEVEL = 2;
```
slifty: Should we define this in one place and use it in both places?

slifty (approved):
How fancy!

nit-level thoughts about the constant name and definition scope, but neither are blocking at all.

PR #777, opened 2026-06-13
Allow numeric strings in metadata

liam-lloyd:
Currently, our metadata validation fails if a field we expect to be a string contains a purely numeric value. This is because we have our XML parser set to parse fields it thinks are numbers as numbers. This commit turns that feature off, because it is legitimate for a string field to have a numeric value. This does mean that when we implement location metadata parsing we'll likely have to convert coordinates and elevations to numbers ourselves before writing to the database.


cecilia-donnelly (approved):
I wonder if there's some way to use the [numberParseOptions](https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/docs/v4%2C%20v5/2.XMLparseOptions.md#numberparseoptions) to handle the coordinates case specifically.

PR #778, opened 2026-06-13
Allow single IPTC keyword in metadata

liam-lloyd:
Currently, we expect the IPTC:Keyword field in metadata files to be an object containing potentially multiple keywords. However, in some files it is simply a single keyword expressed as a string. This commit updates the metadata attacher lambda to handle that case.


cecilia-donnelly (approved):
Is there a reason to have such a long XML file as the fixture? (It seems like a simpler one would be less likely to break for some other reason unrelated to this specific test)

liam-lloyd:
> Is there a reason to have such a long XML file as the fixture? (It seems like a simpler one would be less likely to break for some other reason unrelated to this specific test)

Most of the fixture metadata files used in this package are derived from a real one (from a test environment) in pursuit of authenticity, but it does make sense to a use a smaller one for specific test case like this.

PR #784, opened 2026-06-16
Update queries to use the new name of the record EDTF column

liam-lloyd:
We are renaming record.originalfilecreationtime to record.displaytime, which better reflects its actual meaning. In light of that, this commit updates all stela queries that touch that column to use the new name.


liam-lloyd:
This requires https://github.com/PermanentOrg/back-end/pull/888 (which is why the tests fail)

slifty (approved):

PR #786, opened 2026-06-22
Revert "Bump dompurify from 3.4.10 to 3.4.11"

liam-lloyd:
Reverts PermanentOrg/stela#785

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5907,9 +5907,9 @@
 			}
 		},
 		"node_modules/dompurify": {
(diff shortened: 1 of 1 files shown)


slifty:
Approving but can you share context about what went wrong?

slifty (approved):

liam-lloyd:
> Approving but can you share context about what went wrong?

I should have marked this a draft; this commit lined up timewise with an auth issue we're seeing in dev but it doesn't look like it's actually the cause

PR #787, opened 2026-06-22
Revert "Allow user authentication via SFTP token"

liam-lloyd:
Reverts PermanentOrg/stela#772


slifty:
I believe that this PR was opened as part of the timeout bug / testing if this was the root cause.  It wasn't, so I'm going to close it.

PR #789, opened 2026-06-26
PER-10670: Switch to vitest

liam-lloyd:
Jest doesn't officially support ESM. We'd like to move this repo to ESM,
as that appears to be the way of the future (and we have some
dependencies that in their latest releases only offer ESM support). To
that end, this commit replaces Jest with Vitest.


liam-lloyd:
The codecov difference here seems to come from differences in how jest and vitest count coverage, not from any actual reduction in coverage.

packages/api/src/account/controller/create_storage_adjustments.test.ts
```diff
@@ -1,5 +1,6 @@
 import request from "supertest";
-import { when } from "jest-when";
+import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
```
cecilia-donnelly, line 2: It looks like we imported these from jest in some central location and have to import them in each file from vitest -- is that expected?

cecilia-donnelly (commented):
This looks straightforward and the tests are passing. I have a question about the import setup inline.

cecilia-donnelly:
Some Claude review comments:
```
- packages/event_utils lost its jest.config.js with no vitest.config.ts replacement — dormant since it has no tests yet, but will break  path-alias resolution the moment one is added.
- AGENTS.md still tells contributors to use jest-when, which was removed from package.json in favor of vitest-when.
-  a duplicated hand-rolled @stela/logger mock in trigger_archivematica instead of the new shared mock, vitest-mock-extended added  separately in 6 package.json files instead of hoisted at root, coverage.clean: false risking stale coverage output, and the old silent:   true Jest option having no Vitest equivalent (noisier CI logs).
```

packages/api/src/account/controller/create_storage_adjustments.test.ts
liam-lloyd: Yes, this is a difference in philosophy between jest and vitest. Jest imported a bunch of stuff automatically, vitest wants you to be explicit

liam-lloyd:
> Some Claude review comments:
> 
> ```
> - packages/event_utils lost its jest.config.js with no vitest.config.ts replacement — dormant since it has no tests yet, but will break  path-alias resolution the moment one is added.
> - AGENTS.md still tells contributors to use jest-when, which was removed from package.json in favor of vitest-when.
> -  a duplicated hand-rolled @stela/logger mock in trigger_archivematica instead of the new shared mock, vitest-mock-extended added  separately in 6 package.json files instead of hoisted at root, coverage.clean: false risking stale coverage output, and the old silent:   true Jest option having no Vitest equivalent (noisier CI logs).
> ```

I think it's correct for event_utils to not have a vitest config if it doesn't have tests. I've fixed the other things noted here (except the last one, which doesn't have a fix)

packages/api/src/account/controller/create_storage_adjustments.test.ts
cecilia-donnelly: thank you!

cecilia-donnelly:
Looks like something that changed broke some of the tests unfortunately.

AGENTS.md
```diff
@@ -45,7 +45,7 @@ This is an npm workspace monorepo (workspaces defined in the root `package.json`
 - **Framework:** Express
 - **Database:** PostgreSQL, accessed via TinyPg (SQL file-based query library)
 - **Authentication:** FusionAuth (external identity provider)
-- **Testing:** Jest with ts-jest, Supertest, jest-when, nock, jest-mock-extended
+- **Testing:** Vitest, Supertest, vitest-when, nock, jest-mock-extended
```
cecilia-donnelly: Super tiny, but this still has `jest-mock-extended` in the list, which presumably is now some vitest equivalent.

cecilia-donnelly (approved):
Woohoo!

PR #798, opened 2026-07-06
Clean up emergent dependency problems

liam-lloyd:
It looks like some recent dependency updates (perhaps only in combination with each other?) caused a couple of build errors. This fixes them; see the commit messages for the details of each issue.


cecilia-donnelly (approved):

slifty:
Sorry about that (and also... how...?)

PR #799, opened 2026-07-07
Update Kubernetes to 1.34

liam-lloyd:
Kubernetes 1.33 will soon pass out of standard AWS support, so this commit updates our clusters to 1.34.


slifty (approved):
What's our kubernetes policy / should we be moving up even more?

liam-lloyd:
> What's our kubernetes policy / should we be moving up even more?

We should update more, but EKS requires us to go one version at a time.

PR #801, opened 2026-07-08
Per 10309 fix api best practice violations

liam-lloyd:
This does two small changes to bring our API closer to compliance with our standards
- add an `id` field to record objects (to replace the stuttering `recordId`, though that's retained for now)
- add an `itemType` field to the items returned by GET /folders/{id}/children (which isn't strictly about following the standard but will make the API more friendly to callers

I'm working on other things as part of this ticket as well, but they're bigger changes so I decided to put up this PR separately.

### DIFF
--- a/packages/api/docs/src/models/item.yaml
+++ b/packages/api/docs/src/models/item.yaml
@@ -1,0 +1,20 @@
+item:
+  oneOf:
+    - allOf:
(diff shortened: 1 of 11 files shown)


cecilia-donnelly (approved):
Looks great! Thank you.

PR #802, opened 2026-07-09
Per 10309 add pagination to list endpoints

liam-lloyd:
This is built on top of https://github.com/PermanentOrg/stela/pull/801, which fixes some smaller API standard violations. Only the commits adding pagination to various endpoints are unique to this PR.


cecilia-donnelly (approved):
This looks good to me! Thanks @liam-lloyd .

PR #803, opened 2026-07-10
Remove deprecated storageGiftInMB field from transfer ownership calls

liam-lloyd:
The legacy /archive/transferOwnership endpoint has a storageGiftInMB field, which was historically used to send storage alongside archive ownership transfers. This feature is deprecated, and nothing uses it any longer. Thus this commit removes it from the stela APIs client for the legacy API.


cecilia-donnelly (approved):
Looks good, thank you!

PR #804, opened 2026-07-13
Add timezone field to record and folder locations

liam-lloyd:
The displayTime field of a record or folder tracks the offset associated with that time, but not a specific timezone. This commit adds a timezone field to record and folder locations, since we treat timezone as location metadata (offset is all you need to know to express time accurately; a timezone is a place).


liam-lloyd:
Tests currently fail for lack of https://github.com/PermanentOrg/back-end/pull/897

cecilia-donnelly (approved):
This looks correct to me.

PR #806, opened 2026-07-16
Add documentation for planned dashboard 2030 endpoints

liam-lloyd:
This commit adds the specifications for the new and updated endpoints required to implement the dashboard from our Vision 2030 designs. These endpoints are
- GET /accounts/me
- GET /accounts/me/marketing-tags
- POST /accounts/me/marketing-tags
- POST /archives
- GET /archives/{id}
- PATCH /archive-memberships/{id}
- POST /accounts/me/promo-claim

### DIFF
--- a/packages/api/docs/src/models/archive.yaml
+++ b/packages/api/docs/src/models/archive.yaml
@@ -2,6 +2,7 @@ archive:
   description: Archive object
   type: object
   required:
+    - id
     - archiveId
     - rootFolderId
     - name
@@ -13,8 +14,11 @@ archive:
     - createdAt
     - updatedAt
   properties:
+    id:
+      type: string
     archiveId:
       type: string
+      deprecated: true
     rootFolderId:
       type: string
     description:
@@ -60,8 +64,13 @@ archive:
       $ref: "./archive_membership_role.yaml#/archiveMembershipRole"
     status:
       type: string
+      enum:
+        - ok
+        - orphaned
+        - deleted
+        - generate-avatar
     type:
-      type: string
+      $ref: "./archive_type.yaml#/archiveType"
     createdAt:
       type: string
       format: date-time
@@ -73,3 +82,26 @@ archive:
       enum:
         - chronological
         - reverse_chronological
+archiveSummary:
+  description: A minimal, summary version of an archive object
+  type: object
+  required:
+    - id
+    - name
+    - thumbnailUrls
+  properties:
+    id:
+      type: string
+    name:
+      type: string
+    thumbnailUrls:
+      type: object
+      properties:
+        width200:
+          type: string
+        width500:
+          type: string
+        width1000:
+          type: string
+        width2000:
+          type: string
--- a/packages/api/docs/src/models/archive_membership.yaml
+++ b/packages/api/docs/src/models/archive_membership.yaml
@@ -1,0 +1,18 @@
+archiveMembership:
+  description: A membership relation granting an account some level of access to an archive
+  type: object
+  required: [id, accountId, archive, accessRole, status]
+  properties:
+    id:
+      type: string
+    accountId:
+      type: string
+    archive:
+      $ref: "./archive.yaml#/archiveSummary"
+    accessRole:
+      $ref: "./archive_membership_role.yaml#/archiveMembershipRole"
+    status:
+      type: string
+      enum:
+        - ok
+        - pending
--- a/packages/api/docs/src/models/folder.yaml
+++ b/packages/api/docs/src/models/folder.yaml
@@ -43,7 +43,7 @@ folder:
     shares:
       type: array
       items:
-        $ref: "../models/share.yaml#/share"
+        $ref: "../models/share.yaml#/shareSummary"
     pendingShares:
       nullable: true
       type: array
--- a/packages/api/docs/src/models/share.yaml
+++ b/packages/api/docs/src/models/share.yaml
@@ -1,6 +1,6 @@
-share:
+shareSummary:
   description: >
-    Share Object
+    Share Object Summary
   type: object
   properties:
     id:
--- a/packages/api/docs/src/paths/account.yaml
+++ b/packages/api/docs/src/paths/account.yaml
@@ -149,3 +149,133 @@ storage-adjustment:
         $ref: "../errors.yaml#/404"
       "500":
         $ref: "../errors.yaml#/500"
+accounts/me:
+  get:
+    summary: Get the currently authenticated account.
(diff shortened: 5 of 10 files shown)


liam-lloyd:
For reference:
- The user stories for the Vision 2030 dashboard: https://docs.google.com/document/d/1iKaoE-6wleW23kDk_GmzXfAJwPSRioVXCKK1UJYn6FY/edit?tab=t.0
- Mapping those user stories (excluding the second priority ones and Story 11, which we've elected to defer) to the endpoints they'll need: https://docs.google.com/document/d/11JexdFzrUhR5mVyg2JEgAuoba_MuhCvpu0B1urc_dF4/edit?tab=t.0

packages/api/docs/src/models/archive.yaml
```diff
@@ -13,8 +14,11 @@ archive:
     - createdAt
     - updatedAt
   properties:
+    id:
+      type: string
     archiveId:
       type: string
+      deprecated: true
```
aasandei-vsp, line 21: I am not sure I understand this, it is required but also deprecated. What would be the process to actually get rid of it?

packages/api/docs/src/models/folder.yaml
```diff
@@ -43,7 +43,7 @@ folder:
     shares:
       type: array
       items:
-        $ref: "../models/share.yaml#/share"
+        $ref: "../models/share.yaml#/shareSummary"
```
aasandei-vsp, line 46: Is this a safe change? Could it have a butterfly effect on other endpoints that use the folder model?

packages/api/docs/src/models/share.yaml
```diff
@@ -1,6 +1,6 @@
-share:
+shareSummary:
   description: >
-    Share Object
+    Share Object Summary
```
aasandei-vsp, line 3: Taking into account that only the name changed, but not the properties it exposes, how is this an actual summary? Do we have another share model with more data?

aasandei-vsp (commented):
Please ignore any of the comments if they don't make sense. I am probably missing a lot of context.

packages/api/docs/src/models/archive_membership.yaml
```diff
@@ -0,0 +1,19 @@
...
+    archive:
+      $ref: "./archive.yaml#/archiveSummary"
+    accessRole:
+      $ref: "./archive_membership_role.yaml#/archiveMembershipRole"
+    status:
+      type: string
+      enum:
+        - ok
+        - pending
+        - deleted
```
cecilia-donnelly: I'm trying to think of when we might want to return a deleted membership. Do we notify invitation senders when their invitation is rejected, or members when an owner removes them? I don't think we show revoked/rejected archive memberships anywhere.

packages/api/docs/src/models/folder.yaml
cecilia-donnelly, line 54: Is this a different type than the `shareSummary`?

packages/api/docs/src/models/archive.yaml
liam-lloyd: It's deprecated because we want clients to use `archive.id` instead of `archive.archiveId`, but it's required because we do still promise to include `archiveId` in any archive object we return from the API. Once no clients are using `archive.archiveId` we can remove the field.

packages/api/docs/src/models/share.yaml
liam-lloyd: It's a summary because it's returning summary-level information about a share; we don't presently have any endpoints that return a complete share, but when we do the complete share object would include `createdAt`, `itemId`, `itemType`, possible a whole ArchiveSummary in addition to `archiveId`, and possibly a couple other fields.

packages/api/docs/src/paths/account.yaml
```diff
@@ -149,3 +149,133 @@ storage-adjustment:
...
+                  type: array
+                  items:
+                    type: string
+      "401":
+        $ref: "../errors.yaml#/401"
+      "400":
+        $ref: "../errors.yaml#/400"
+      "500":
+        $ref: "../errors.yaml#/500"
+  get:
```
cecilia-donnelly, line 215: I take it this is so the member can see their answers to the widgets on the dashboard? Do those disappear after being filled in, though?

cecilia-donnelly: What I'm really asking is whether or not we need a `GET` here.

packages/api/docs/src/paths/account.yaml
```diff
@@ -149,3 +149,133 @@ storage-adjustment:
...
+          schema:
+            type: object
+            required: [promoCode]
+            properties:
+              promoCode:
+                type: string
+    responses:
+      "200":
+        description: >
+          The amount of storage granted by the promo code
```
cecilia-donnelly: In GB?

packages/api/docs/src/paths/archive_membership.yaml
```diff
@@ -0,0 +1,50 @@
...
+              type: object
+              properties:
+                data:
+                  $ref: "../models/archive_membership.yaml#/archiveMembership"
+      "400":
+        $ref: "../errors.yaml#/400"
+      "401":
+        $ref: "../errors.yaml#/401"
+      "500":
+        $ref: "../errors.yaml#/500"
```
cecilia-donnelly, line 50: I take it that creating a new membership (that is, an invitation to an archive) isn't part of the dashboard so doesn't get included here, but would be a POST on this endpoint?

cecilia-donnelly (commented):
Exciting!

packages/api/docs/src/paths/account.yaml
cecilia-donnelly: Yes, these will persist (pending more product review) because we already have a way to edit the tags. "Persistence/editability is TBD based on whether editing answers is already supported by the backend. If editing is not already supported, the widget does not persist after answering and we will not build new editing functionality for this pass."

packages/api/docs/src/models/archive_membership.yaml
cecilia-donnelly: @liam-lloyd and I think we can remove this.

packages/api/docs/src/paths/account.yaml
cecilia-donnelly: @liam-lloyd will add a unit here.

cecilia-donnelly (approved):
Talked to @liam-lloyd and he will make the small tweaks mentioned. Hurray!

PR #808, opened 2026-07-20
Switch to ESM modules

liam-lloyd:
ESM, rather than CommonJS, seems to be the module system of the future. We have some dependencies that we can't update without adopting ESM. So, this commit updates this project to use ESM instead of CommonJS.


liam-lloyd:
(I used this ticket to test out auto-mode Claude Code running in a Docker Sandbox)

cecilia-donnelly (approved):
This is simultaneously very big and very limited, so I did the random sampling approach to my review. Tests are passing so I believe them that all is well!

PR #809, opened 2026-07-20
Fix @stela/publisher-utils tests

liam-lloyd:
In the recent switch from jest to vitest, we broke our publisher-utils tests. We didn't catch this because our CI wasn't configured to run these tests. This commit fixes the tests and updates the CI to test this package in the future.


cecilia-donnelly (approved):
Great, thanks for catching this @liam-lloyd !

PR #813, opened 2026-07-23
Add GET /accounts/me/marketing-tags endpoint

liam-lloyd:
The Vision 2030 dashboard design contains the possibility of displaying a users selected goals on the dashboard. To support this, this commit adds a GET /accounts/me/marketing-tags endpoint, which will retrieve those values from Mailchimp.

### DIFF
--- a/packages/api/src/account/service.ts
+++ b/packages/api/src/account/service.ts
@@ -1,6 +1,8 @@
 import { Md5 } from "ts-md5";
 import createError from "http-errors";
+import { HTTP_STATUS } from "@pdc/http-status-codes";
 import { logger } from "@stela/logger";
+import type { ErrorResponse } from "@mailchimp/mailchimp_marketing";
 
 import { db } from "../database.js";
 import { MailchimpMarketing } from "../mailchimp.js";
@@ -10,6 +12,7 @@ import type { CreateEventRequest } from "../event/models.js";
 
 import {
 	type UpdateTagsRequest,
+	type GetMarketingTagsResponse,
 	type SignupDetails,
 	type GetAccountArchiveResult,
 	type LeaveArchiveRequest,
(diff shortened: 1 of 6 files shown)


packages/api/src/account/service.ts
```diff
@@ -151,6 +154,41 @@ const updateTags = async (requestBody: UpdateTagsRequest): Promise<void> => {
...
+	status: number;
+	response?: {
+		body: ErrorResponse;
+	};
+}
+
+const isMailchimpApiError = (err: unknown): err is MailchimpApiError =>
+	err instanceof Object &&
+	"status" in err &&
+	typeof (err as { status: unknown }).status === "number";
```
cecilia-donnelly, line 167: Why can't we check against the `ErrorResponse` type?

packages/api/src/account/service.ts
```diff
@@ -151,6 +154,41 @@ const updateTags = async (requestBody: UpdateTagsRequest): Promise<void> => {
...
+	try {
+		const response = await MailchimpMarketing.lists.getListMemberTags(
+			process.env["MAILCHIMP_COMMUNITY_LIST_ID"] ?? "",
+			Md5.hashStr(requestBody.emailFromAuthToken),
+		);
+		return { items: response.tags.map((tag) => tag.name) };
+	} catch (err) {
+		if (isMailchimpApiError(err)) {
+			if (err.status === HTTP_STATUS.CLIENT_ERROR.NOT_FOUND.valueOf()) {
+				return { items: [] };
```
cecilia-donnelly, line 182: Nice, this is what I would expect.

cecilia-donnelly (commented):
This all makes sense! I got an error back from Mailchimp locally (as expected), hurray. I don't quite understand the `isMailchimpApiError`, which I asked about inline.

packages/api/src/account/service.ts
liam-lloyd: Because this logic is executing at runtime, where Typescript interfaces don't exist.

cecilia-donnelly (approved):

PR #816, opened 2026-07-23
Fix broken BadRequest import

liam-lloyd:
Currently, we have one file that attempts to import BadRequest directly from http-errors. Because http-errors is a CommonJS module, this doesn't work (though it fails in a way that the Typescript compiler doesn't catch). This commit fixes the problem but just importing createError and accessing BadRequest through that.


cecilia-donnelly (approved):
Thanks for catching this!

PR #817, opened 2026-07-24
Add GET /accounts/me endpoint

liam-lloyd:
The 2030 dashboard will require various information about the logged in user's account. To serve those use cases, this commit adds a GET /accounts/me endpoint that returns that account, including its archive memberships.

### DIFF
--- a/.sqlfluff
+++ b/.sqlfluff
@@ -14,7 +14,7 @@ param_style = colon
 spacing_within = touch:inline
 
 [sqlfluff:rules:references.keywords]
-ignore_words = name, owner, type, trigger, shares, location, action, version, position, precision
+ignore_words = name, owner, type, trigger, shares, location, action, version, position, precision, settings
 
 [sqlfluff:rules:capitalisation.functions]
 extended_capitalisation_policy = upper
--- a/packages/api/docs/src/models/account.yaml
+++ b/packages/api/docs/src/models/account.yaml
@@ -40,7 +40,7 @@ notificationTypesEnabled:
 account:
   description: An object representing a user account on permanent.org
   type: object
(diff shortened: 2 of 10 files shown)


packages/api/docs/src/models/account.yaml
```diff
@@ -64,6 +65,7 @@ account:
       type: string
     defaultArchiveId:
       type: string
+      nullable: true
     address:
```
cecilia-donnelly, line 69: This should be nullable too, right?

cecilia-donnelly: I see, the individual properties can all be null but the overall address cannot. Is that right?

packages/api/src/account/models.ts
```diff
@@ -56,27 +71,13 @@ export interface Account {
...
-	zip: string | null;
-	country: string | null;
-	hideChecklist: boolean;
-	allowSftpDeletion: boolean;
-	notificationPreferences: RawNotificationPreferences;
-	status: AccountStatus;
-	type: AccountType;
+export interface AccountRow extends Account {
+	totalPages: number;
 }
```
cecilia-donnelly, line 81: This is a nice way to do it.

.sqlfluff
```diff
@@ -14,7 +14,7 @@ param_style = colon
 spacing_within = touch:inline
 
 [sqlfluff:rules:references.keywords]
-ignore_words = name, owner, type, trigger, shares, location, action, version, position, precision
+ignore_words = name, owner, type, trigger, shares, location, action, version, position, precision, settings
```
cecilia-donnelly, line 17: ah, well.

cecilia-donnelly (approved):

PR #818, opened 2026-07-24
Add POST /accounts/me/promo-claim endpoint

liam-lloyd:
Users will be able to claim their welcome promo from the dashboard in the Vision 2030 design. To that end, this commit adds a promo-claim endpoint.

### DIFF
--- a/packages/api/docs/src/paths/account.yaml
+++ b/packages/api/docs/src/paths/account.yaml
@@ -245,7 +245,6 @@ accounts/me/promo-claim:
     tags:
       - accounts
       - promo
-      - unimplemented
     requestBody:
       content:
         application/json:
@@ -269,11 +268,13 @@ accounts/me/promo-claim:
                   required: [storageGrantInGb]
                   properties:
                     storageGrantInGb:
-                      type: integer
+                      type: number
       "401":
         $ref: "../errors.yaml#/401"
       "403":
         $ref: "../errors.yaml#/403"
+      "404":
+        $ref: "../errors.yaml#/404"
       "400":
         $ref: "../errors.yaml#/400"
       "500":
--- a/packages/api/src/promo/models.ts
+++ b/packages/api/src/promo/models.ts
@@ -1,3 +1,26 @@
+export interface ClaimPromoRequest {
+	emailFromAuthToken: string;
+	userSubjectFromAuthToken: string;
+	promoCode: string;
+}
+
+export interface ClaimPromoResponse {
+	storageGrantInGb: number;
+}
+
+export interface PromoByCodeRow {
+	id: string;
+	storageInMB: string;
+	expirationTimestamp: Date;
+	remainingUses: string;
+	status: string;
+}
+
+export interface AccountPromoCheckRow {
+	accountId: string;
+	existingClaimId: string | null;
+}
+
 export interface CreatePromoRequest {
 	emailFromAuthToken: string;
 	adminSubjectFromAuthToken: string;
@@ -11,22 +34,22 @@ export interface Promo {
 	id: string;
 	code: string;
 	storageInMB: number;
-	expirationTimestamp: string;
+	expirationTimestamp: Date;
 	remainingUses: number;
 	status: string;
 	type: string;
-	createdAt: string;
-	updatedAt: string;
+	createdAt: Date;
+	updatedAt: Date;
 }
 
 export interface PromoRow {
 	id: string;
 	code: string;
 	storageInMB: string;
-	expirationTimestamp: string;
+	expirationTimestamp: Date;
 	remainingUses: string;
 	status: string;
 	type: string;
-	createdAt: string;
-	updatedAt: string;
+	createdAt: Date;
+	updatedAt: Date;
 }
--- a/packages/api/src/promo/service.ts
+++ b/packages/api/src/promo/service.ts
@@ -1,7 +1,116 @@
 import createError from "http-errors";
 import { logger } from "@stela/logger";
 import { db } from "../database.js";
-import type { CreatePromoRequest, Promo, PromoRow } from "./models.js";
+import { KB } from "../constants.js";
+import type {
+	AccountPromoCheckRow,
+	ClaimPromoRequest,
+	ClaimPromoResponse,
+	CreatePromoRequest,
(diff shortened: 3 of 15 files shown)


packages/api/docs/src/paths/account.yaml
```diff
@@ -246,7 +246,6 @@ accounts/me/promo-claim:
     tags:
       - accounts
       - promo
-      - unimplemented
     requestBody:
```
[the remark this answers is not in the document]
slifty: And here I was thinking robots couldn't do math any more.

packages/api/src/promo/service.ts
```diff
@@ -1,7 +1,112 @@
...
+			throw new createError.BadRequest("Invalid promo code");
+		}
+		if (promo.status !== "status.promo.valid") {
+			throw new createError.BadRequest("Invalid promo code");
+		}
+		if (new Date(promo.expirationTimestamp) <= new Date()) {
+			throw new createError.BadRequest("Invalid promo code");
+		}
+		if (+promo.remainingUses <= 0) {
+			throw new createError.BadRequest("Invalid promo code");
```
slifty, line 38: This is repeated 4x -- could we use a validator for these checks with a single error throw (or a single if with `||` clauses and a single error block)

packages/api/src/promo/service.ts
```diff
@@ -1,7 +1,112 @@
...
+
+		const {
+			rows: [accountCheck],
+		} = accountCheckResult;
+		if (accountCheck === undefined) {
+			throw new createError.NotFound("Account not found");
+		}
+		if (accountCheck.existingClaimId !== null) {
+			throw new createError.BadRequest(
+				"Promo code has already been claimed by this account",
```
slifty, line 61: I was curious if there was a lurking race condition here in which someone could still claim twice -- apparently because we use `FOR UPDATE` in the sql we're safe.

That said, we could be even safer by adding a UNIQUE constraint `account_promo (accountid, promoid)` which would prevent the same account from using a promo twice at the schema level rather than the update.  Maybe there's a tradeoff / reason not to!

Also, should we have a unique constraint on `promo.code` / right now could we have multiple promos with the same code?  Maybe we should fix that too 😅 (in a separate PR I imagine)

packages/api/src/storage/queries/adjust_account_storage.sql
slifty, line 9: If this row doesn't exist should we throw a DB error for the sake of defensiveness?

packages/api/src/promo/service.ts
```diff
@@ -1,7 +1,112 @@
...
+			throw new createError.BadRequest("Invalid promo code");
+		}
+		if (new Date(promo.expirationTimestamp) <= new Date()) {
+			throw new createError.BadRequest("Invalid promo code");
+		}
+		if (+promo.remainingUses <= 0) {
+			throw new createError.BadRequest("Invalid promo code");
+		}
+
+		const accountCheckResult = await transactionDb
```
slifty, line 41: Account storage adjustment will use the promo code but fail to add account space if there is no account_space row -- we should check that here too (and we can write a test to cover that case / use `test+3@permanent.org`

packages/api/src/promo/service.ts
```diff
@@ -1,7 +1,112 @@
...
+				throw new createError.InternalServerError(
+					"Failed to check promo claim status",
+				);
+			});
+
+		const {
+			rows: [accountCheck],
+		} = accountCheckResult;
+		if (accountCheck === undefined) {
+			throw new createError.NotFound("Account not found");
```
slifty, line 57: This 404 possibility should be added to the spec

packages/api/src/promo/models.ts
```diff
@@ -1,3 +1,26 @@
...
+}
+
+export interface ClaimPromoResponse {
+	storageGrantInGb: number;
+}
+
+export interface PromoByCodeRow {
+	id: string;
+	storageInMB: string;
+	expirationTimestamp: string;
```
slifty: This is a timestamp in the DB -- (Claude seems to think that means it would be returned as a `Date` rather than a `string`)

slifty (changes requested):
Since storage is so precious I think it's worth checking for the potential "no account_storage row" edge case

packages/api/src/promo/service.ts
liam-lloyd: Those do seem like good DB constraints to add

packages/api/src/storage/queries/adjust_account_storage.sql
liam-lloyd: Postgres won't let us throw an error directly from the query, but we should definitely check that this returns a row and throw if it doesn't

packages/api/src/promo/service.ts
liam-lloyd: An account without an `account_space` row is a decidedly broken state; I think if we handle it at the adjust storage query (by throwing 500 and rolling back the transaction) we don't need to check for it in advance.

packages/api/src/storage/queries/adjust_account_storage.sql
slifty: Got ya -- I did mean something like this in the `.sql` (not necessarily the query iteself), which may not actually be a good idea:

https://github.com/PhilanthropyDataCommons/service/blob/ad999b293d8fdc427654af514e2076dcf9588ba5/src/database/initialization/1_assert_changemaker_field_value_not_forbidden.sql#L17-L21

```
		RAISE EXCEPTION
			'Refusing to serialize forbidden changemaker_field_value (%)',
			changemaker_field_value.id
			USING ERRCODE = '22023'; -- invalid_parameter_value
```

liam-lloyd: Yeah, I'm hesitant to introduce plpgsql into the query files (this isn't a reasoned-out hesitancy though, so I could be convinced, but in this case checking whether any rows were updated seems just as good)

slifty: Seems reasonable to keep it out / stick with TS validation!

slifty (approved):
Looks good!  Thanks for the iteration.

PR #819, opened 2026-07-27
Add PATCH /archive-memberships endpoint

liam-lloyd:
From the 2030 dashboard, a user should be able to accept an invitation to join an archive. This requires changing the status of an archive membership object from "pending" to "ok". To that end, this commit adds a PATCH /archive-memberships endpoint that can be used to update archive membership objects.


packages/api/src/archive_membership/validators.ts
```diff
@@ -0,0 +1,36 @@
...
+	data: unknown,
+): asserts data is UpdateArchiveMembershipRequest => {
+	const validation = Joi.object()
+		.keys({
+			...fieldsFromUserAuthentication,
+			accessRole: Joi.string()
+				.valid("viewer", "contributor", "editor", "curator", "manager")
+				.optional(),
+			status: Joi.string().valid("ok").optional(),
+		})
```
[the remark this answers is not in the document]
slifty: I agree with this sentiment / avoiding hand written strings when possible.

packages/api/src/archive_membership/controller.test.ts
```diff
@@ -0,0 +1,272 @@
...
+import { app } from "../app.js";
+import { verifyUserAuthentication } from "../middleware/index.js";
+import { db } from "../database.js";
+import type { ArchiveMembership } from "./models.js";
+import { mockVerifyUserAuthentication } from "../../test/middleware_mocks.js";
+import { mockSqlCall } from "../../test/mock_sql.js";
+
+vi.mock("../database");
+vi.mock("../middleware");
+
```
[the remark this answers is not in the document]
liam-lloyd: The mock the logger to test logger calls, not to stop it from logging

packages/api/docs/src/paths/archive_membership.yaml
slifty: We can take this tag out

slifty: This doc block is inaccurate now (since we removed `deleted`)

packages/api/src/archive_membership/service.ts
```diff
@@ -0,0 +1,137 @@
...
+	UpdateArchiveMembershipRequest,
+} from "./models.js";
+
+interface PermissionsDataRow {
+	archiveId: string;
+	accountEmail: string;
+	isOwner: boolean;
+}
+
+const getCallerAccessRoleOr404 = async (
```
slifty: Claude pointed out that this is almost identical to `getArchiveAccessRole` -- may be worth some kind of restructuring to avoid duplication

packages/api/src/archive_membership/queries/get_archive_membership.sql
```diff
@@ -0,0 +1,41 @@
...
+INNER JOIN
+  account ON account_archive.accountid = account.accountid
+INNER JOIN
+  archive ON account_archive.archiveid = archive.archiveid
+INNER JOIN
+  profile_item
+  ON
+    archive.archiveid = profile_item.archiveid
+    AND profile_item.fieldnameui = 'profile.basic'
+    AND profile_item.status != 'status.generic.deleted'
```
slifty: Claude flagged this saying that most places use `status = 'status.generic.ok' AND profile_item.string1 IS NOT NULL` instead of != deleted.

This implementation means:

1. An archive that doesn't have an un-deleted profile.basic row drops out of the INNER JOIN.  Claude said that this means the update succeeds but the request returns 500 "Failed to retrieve updated archive membership"

2. If there's more than one non-deleted `profile.basic` row there will be multiple result rows (I don't know if this can happen?)

3. A null string1 yields name: null, which violates `archiveSummary`'s required: `[name] / type: string`.

packages/api/src/archive_membership/controller.test.ts
```diff
@@ -0,0 +1,272 @@
...
+	test("should return 404 if a pending member tries to update another user's membership", async () => {
+		mockVerifyUserAuthentication(
+			"test+1@permanent.org",
+			"bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb",
+		);
+		await agent
+			.patch("/api/v2/archive-memberships/1")
+			.send({ accessRole: "contributor" })
+			.expect(404);
```
slifty, line 114: We should make sure the api spec documents 404 as a possibility.

packages/api/src/archive_membership/controller.test.ts
```diff
@@ -0,0 +1,272 @@
...
+
+	test("should return 403 if a non-manager tries to update accessRole", async () => {
+		mockVerifyUserAuthentication(
+			"test+3@permanent.org",
+			"dddddddd-dddd-dddd-dddd-dddddddddddd",
+		);
+		await agent
+			.patch("/api/v2/archive-memberships/2")
+			.send({ accessRole: "contributor" })
+			.expect(403);
```
slifty, line 125: Likewise we should make sure the api spec documents 403 as a possibility.

slifty (commented):
A few things that I found including some claude assisted finds!

Claude also mentioned that we don't log an audit event here; Do we have any documented policies around when to log events?

liam-lloyd:
> A few things that I found including some claude assisted finds!
> 
> Claude also mentioned that we don't log an audit event here; Do we have any documented policies around when to log events?

Oh, this should definitely have an event, good call. Any writes to the database should get events (I'm not sure that's written down but it is the intent)

packages/api/src/archive_membership/queries/get_archive_membership.sql
liam-lloyd: It should in practice be impossible for an archive to not have precisely one `profile.basic` row with a non-null `string1`. I'll add 500 errors for multiple rows or a null archive name, since this would be considered corrupted data.

liam-lloyd:
Needs https://github.com/PermanentOrg/back-end/pull/904 to fix the tests

PR #821, opened 2026-07-31
Optimize API tests

liam-lloyd:
This commit makes two changes that make the API tests interact with the database more efficiently. The first is that it updates the tests to use a single database call to set up all their fixture data, rather than using one call per affected table. The second is that it removes redundant database clears after tests (because tests all clear the database before running). This speeds up the API tests by ~25% on my machine.


cecilia-donnelly (approved):
Nice!

PR #834, opened 2026-08-05
Remove redundant terraform plan steps in deploys

liam-lloyd:
Currently when we deploy this project we run `terraform plan`, then `terraform apply`. The intent was that if there was something wrong we'd catch it at the plan stage without ever getting to the apply. However, this was based on a misunderstanding of terraform; `terraform apply` runs a plan first, then applies it. Thus errors that can be caught at planning will still be caught before any changes take place even if we just run `terraform apply`.

### DIFF
--- a/.github/workflows/dev_deploy.yml
+++ b/.github/workflows/dev_deploy.yml
@@ -50,23 +50,6 @@ jobs:
       - name: Terraform Validate
         id: validate
         run: terraform validate -no-color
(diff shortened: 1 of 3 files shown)


liam-lloyd:
This helps with our timeout problem by reducing the number of opportunities available for terraform to time out.

cecilia-donnelly (approved):

PR #835, opened 2026-08-06
Add accessRole field to records and folders

liam-lloyd:
When clients access records and folders, they need to know what permissions level the caller has with respect to that record or folder. To that end, this commit adds an accessRole field to record and folder objects returned by the API.

### DIFF
--- a/packages/api/src/access/permission.ts
+++ b/packages/api/src/access/permission.ts
@@ -1,6 +1,10 @@
 import createError from "http-errors";
 import { logger } from "@stela/logger";
-import { AccessRole } from "./models.js";
+import {
+	AccessRole,
+	type ArchiveMembershipRole,
+	accessRoleToArchiveMembershipRole,
+} from "./models.js";
 import { db } from "../database.js";
 
 const VIEWER_ACCESS_ROLE_RANK = 1;
@@ -25,17 +29,65 @@ export const accessRoleLessThan = (
 	return accessRoleRank.get(roleOne) < accessRoleRank.get(roleTwo);
 };
 
-const moreLimitedAccessRole = (
-	roleOne: AccessRole | null,
-	roleTwo: AccessRole | null,
+export const leastPermissiveAccessRole = (
+	roleOne: AccessRole | null | undefined,
+	roleTwo: AccessRole | null | undefined,
 ): AccessRole | null => {
-	if (roleOne === null) {
-		return roleTwo;
+	if (roleOne === null || roleOne === undefined) {
+		return roleTwo ?? null;
 	}
-	if (roleTwo === null) {
+	if (roleTwo === null || roleTwo === undefined) {
 		return roleOne;
 	}
 	return accessRoleLessThan(roleOne, roleTwo) ? roleOne : roleTwo;
+};
+
+export const mostPermissiveAccessRole = (
+	roles: Array<AccessRole | null | undefined>,
+): AccessRole | null =>
+	roles.reduce<AccessRole | null>((accumulator, role) => {
+		if (role === null || role === undefined) {
+			return accumulator;
+		}
+		if (accumulator === null) {
+			return role;
+		}
+		return accessRoleLessThan(accumulator, role) ? role : accumulator;
+	}, null);
+
+export interface ShareAccessRolePair {
+	archiveAccessRole: AccessRole;
+	shareAccessRole: AccessRole;
+}
+
+export const resolveAccessRole = (input: {
+	archiveAccessRole: AccessRole | null;
+	shareAccessRoles: ShareAccessRolePair[] | null;
+	shareTokenGrantsAccess: boolean;
+	isPublic: boolean;
+}): ArchiveMembershipRole => {
+	const roles: Array<AccessRole | null> = [
+		input.archiveAccessRole,
+		...(input.shareAccessRoles ?? []).map((share) =>
+			leastPermissiveAccessRole(share.archiveAccessRole, share.shareAccessRole),
+		),
+	];
+	if (input.shareTokenGrantsAccess) {
+		roles.push(AccessRole.Viewer);
+	}
+	if (input.isPublic) {
+		roles.push(AccessRole.Viewer);
+	}
+
+	const accessRole = mostPermissiveAccessRole(roles);
+	if (accessRole === null) {
+		// Should be unreachable: the SQL WHERE clause that selects a row already
+		// guarantees at least one access path applies.
+		throw createError.InternalServerError(
+			"Unable to resolve caller's access role for item",
+		);
+	}
+	return accessRoleToArchiveMembershipRole(accessRole);
 };
 
 export const getItemAccessRole = async (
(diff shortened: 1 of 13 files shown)


packages/api/src/access/permission.ts
```diff
@@ -25,19 +29,67 @@ export const accessRoleLessThan = (
...
+export const resolveAccessRole = (input: {
+	archiveAccessRole: AccessRole | null;
+	shareAccessRoles: ShareAccessRolePair[] | null;
+	shareTokenGrantsAccess: boolean;
+	isPublic: boolean;
+}): ArchiveMembershipRole => {
+	const roles: Array<AccessRole | null> = [
+		input.archiveAccessRole,
+		...(input.shareAccessRoles ?? []).map((share) =>
+			leastPermissiveAccessRole(share.archiveAccessRole, share.shareAccessRole),
```
slifty, line 70: Claude flagged this line: `A null accessrole on a share removes the cap instead of denying`

It's saying that this line "caps" a share at the caller's own archive role via `leastPermissiveAccessRole`. But apparently this helper treats nullish as "use the other" instead of returning null.

It's saying this means a share row where `accessrole IS NULL` pointing at an archive the caller owns resolves to `accessRole: "owner"` on a folder in someone else's archive.

Basically, missing / null should mean "no access" in this case.

packages/api/src/access/permission.ts
```diff
@@ -25,19 +29,67 @@ export const accessRoleLessThan = (
...
+
+	const accessRole = mostPermissiveAccessRole(roles);
+	if (accessRole === null) {
+		// Should be unreachable: the SQL WHERE clause that selects a row already
+		// guarantees at least one access path applies.
+		throw createError.InternalServerError(
+			"Unable to resolve caller's access role for item",
+		);
+	}
+	return accessRoleToArchiveMembershipRole(accessRole);
```
slifty, line 88: We actually have a little bug in our implementation of `accessRoleToArchiveMembershipRole` which claude flagged -- the way it's defined *can return undefined* if the `accessRole` doesn't map to an enum value.

TS treats array access as never returning undefined -- that's not accurate here, and could mean we return undefined at runtime even though TS doesn't account for that.

We should harden that implementation / cover the undefined case.

packages/api/src/access/permission.ts
```diff
@@ -25,19 +29,67 @@ export const accessRoleLessThan = (
...
+	if (input.shareTokenGrantsAccess) {
+		roles.push(AccessRole.Viewer);
+	}
+	if (input.isPublic) {
+		roles.push(AccessRole.Viewer);
+	}
+
+	const accessRole = mostPermissiveAccessRole(roles);
+	if (accessRole === null) {
+		// Should be unreachable: the SQL WHERE clause that selects a row already
```
slifty, line 82: Claude flagged that this comment isn't true... I have to admit I don't fully grok why... but it sounded credible enough so I'm going to be rude and post claude's comment verbatim:

```
1. Visibility and role resolution are two independent copies of the access rules, and they don't agree

  resolveAccessRole throws a 500 when no path applies, justified by packages/api/src/access/permission.ts:84:

  ▎ // Should be unreachable: the SQL WHERE clause that selects a row already guarantees at least one access path applies.

  That invariant isn't actually enforced — the new subqueries re-derive the access rules with different predicates than the WHERE clause that admits the row. Where they disagree, the
  "unreachable" branch is reachable and the whole page 500s.

  get_folders.sql — visibility via account_by_share (packages/api/src/folder/queries/get_folders.sql:448) is gated by access.status != 'status.generic.deleted' (:210) and folder_link.status != 
  'status.generic.deleted' (:393). The new shareAccessRoles subquery requires access.status = 'status.generic.ok' and share_folder_link.status = 'status.generic.ok'. status.generic.pending /
  invited / processing are all live values in this codebase, and access.status is nullable text with no constraint (database/base.sql:27). Any such row → visible folder, NULL share roles, 500.

  Same query, membership path: account_by_archive (:438) never checks account.status; the new archiveAccessRole subquery adds AND account.status = 'status.auth.ok'. Your own fixture has an
  account in exactly that state — account 5 / test+3@permanent.org is status.generic.invited (packages/api/src/folder/fixtures/create_test_accounts.sql) and is an ok-status owner of archive 6
  (create_test_account_archives.sql, row id 36). A request from that account for a folder in archive 6 is visible-but-unresolvable.

  get_records.sql — the same two divergences: record_account / share_account in the WHERE (packages/api/src/record/queries/get_records.sql:407-408) have no account.status filter, and the outer
  folder_link join uses != 'status.generic.deleted' (:374) against the subquery's = 'status.generic.ok'.

  The public and share-token paths do match exactly, so those are fine.

  The fix I'd push for: compute the roles in a CTE and drive the WHERE clause off that same CTE, so visibility and role are one expression rather than two that must be kept in sync by hand. Then
  the throw is genuinely unreachable. Short of that, the predicates need to be made character-for-character identical — and one of the divergences is a real behavior question you should decide
  deliberately rather than inherit: should an account whose status isn't auth.ok see these items at all? Right now the two halves of the code disagree about the answer.
```

packages/api/src/folder/queries/get_folders.sql
```diff
@@ -325,7 +325,52 @@ all_folders AS (
       folder.thumburl2000,
       '256',
       folder.thumbnail256
-    ) AS "thumbnailUrls"
+    ) AS "thumbnailUrls",
+    (
+      SELECT account_archive.accessrole
```
slifty, line 331: Claude flagged that we don't have a unique constraint on `account_archive(accountid, 
  archiveid)` and so this subquery could return more than one result => this would then throw a `Two ok-status rows for the same pair and Postgres raises 21000 more than one row returned by a subquery used as an expression`

It suggests adding an `ORDER BY ... LIMIT 1` clause to resolve the potential edge case.

packages/api/src/access/permission.ts
```diff
@@ -25,19 +29,67 @@ export const accessRoleLessThan = (
...
+		// guarantees at least one access path applies.
+		throw createError.InternalServerError(
+			"Unable to resolve caller's access role for item",
```
slifty, line 85: Claude suggested including the item ID in the error her so if it does somehow happen we'd be able to more easily diagnose.  Sounds reasonable to me!

packages/api/src/access/permission.ts
liam-lloyd: Claude's catching an extreme edge case here, because `access` doesn't require that `accessrole` is non-nullable. I'll fix that rather than change the logic here (`leastPermissiveAccessRole` has uses elsewhere where its null handling is desired, and if we do things correctly nulls will be impossible here)

packages/api/src/access/permission.ts
liam-lloyd: We should be protected here by Typescript's type checking; that shouldn't allow a non-`AccessRole` argument to this function

packages/api/src/access/permission.ts
liam-lloyd: The problem Claude is seeing here isn't real because both `folder_link` and `access` have only the two statuses. But it does seem like a good idea to make these checks consistent.

packages/api/src/access/permission.ts
liam-lloyd: There's not another reason to pass itemId to this function though, and this is supposed to be unreachable

packages/api/src/access/permission.ts
slifty: I very much approve of that approach!  marking this as resolved.

packages/api/src/access/permission.ts
slifty: Ahhh yes right!

So the "risk" is that if we're pulling `AccessRole` from a sql query that is type cast (e.g. at the query we say "the shape of this data is `AccessRole`") then the risk is that the query might return something that violates that contract."

This is a root truth / risk of the DB integration point, and that would indeed just be a schema bug not an issue for us.

Carry on / sorry about that!

PR #836, opened 2026-08-06
Add GET /archives/{id} endpoint

liam-lloyd:
This commit adds an endpoint for retrieving an archive by its ID to support the Vision 2030 dashboard.

### DIFF
--- a/packages/api/src/archive/queries/get_archives.sql
+++ b/packages/api/src/archive/queries/get_archives.sql
@@ -1,48 +1,38 @@
 WITH all_archives AS (
   SELECT
+    archive.archiveid AS id,
     archive.archiveid AS "archiveId",
-    archive.archivenbr AS "archiveNbr",
     basic_profile_item.string1 AS name,
     text_data.valuetext AS description,
     archive.public,
     archive.publicdt AS "publicAt",
     archive.allowpublicdownload AS "allowPublicDownload",
+    archive.milestonesortorder AS "milestoneSortOrder",
+    archive.createddt AS "createdAt",
+    archive.updateddt AS "updatedAt",
+    root_folder.folderid AS "rootFolderId",
+    archive.payeraccountid AS "payerAccountId",
+    user_access.accessrole AS "callerMembershipRole",
+    CASE
+      WHEN archive.status = 'status.generic.orphaned' THEN 'orphaned'
+      WHEN archive.status = 'status.archive.gen_avatar' THEN 'generate-avatar'
+      WHEN archive.status IN ('status.generic.ok', 'status.auth.ok') THEN 'ok'
+    END AS status,
+    CASE
+      WHEN archive.type = 'type.archive.person' THEN 'person'
+      WHEN archive.type = 'type.archive.family' THEN 'group'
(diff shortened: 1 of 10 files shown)


packages/api/src/archive/queries/get_archives.sql
```diff
@@ -0,0 +1,124 @@
...
+      WHEN archive.status = 'status.archive.gen_avatar' THEN 'generate-avatar'
+      WHEN archive.status IN ('status.generic.ok', 'status.auth.ok') THEN 'ok'
+    END AS status,
+    CASE
+      WHEN archive.type = 'type.archive.person' THEN 'person'
+      WHEN archive.type = 'type.archive.family' THEN 'group'
+      WHEN archive.type IN (
+        'type.archive.organization',
+        'type.archive.nonprofit'
+      ) THEN 'organization'
```
cecilia-donnelly: I take it this because in the billing/upload code we check the type directly from the database, not from the client, so the client never needs to make this distinction.

packages/api/src/archive/queries/search_archives.sql
```diff
@@ -56,7 +56,7 @@ WITH all_archives AS (
     ON
       archive.archiveid = description_profile_item.archiveid
       AND description_profile_item.fieldnameui = 'profile.description'
-      AND basic_profile_item.status != 'status.generic.deleted'
+      AND description_profile_item.status != 'status.generic.deleted'
```
cecilia-donnelly, line 59: Is this a separate bugfix?

packages/api/src/archive/service/get_archives.ts
```diff
@@ -0,0 +1,78 @@
...
+							{
+								archiveIds: requestQuery.archiveIds,
+								pageSize: requestQuery.pageSize,
+							},
+							nextCursor,
+						),
+			totalPages: result.rows[0]?.totalPages ?? 0,
+		},
+	};
+};
```
cecilia-donnelly, line 78: This seems to belong with a different endpoint (`archives` GET) and should perhaps be a separate commit? Or maybe I'm misreading?

packages/api/src/archive/queries/search_archives.sql
liam-lloyd: Yes; I'll move it to its own commit

packages/api/src/archive/service/get_archives.ts
liam-lloyd: I wanted to implement `GET /archives/{id}` as a special case of `GET /archives` the same way we've done for `GET /records/{id}` and `GET /folders/{id}`. The point of this is consistency, and to make it extremely easy to expose `GET /archives` once we need it so we're not tempted to make repeated calls to `GET /archives/{id}` as a shortcut.

packages/api/src/archive/queries/get_archives.sql
liam-lloyd: I was actually thinking about how we're planning to drop the nonprofit type, but on reconsideration the client probably is still using it in available storage calculations, so I will expose it here.

